PR #18, opened 2021-03-18
NTBS-NONE Use per-developer databases

benjimarshall:
### Description
Update the README, add support for a per-developer publishing script in the `.sqlproj` file, and ignore this script in source control.

### Testing
No formal testing required, though this is based on me going through the process of setting up a new reporting database with Nancy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,10 @@ To make a new copy of the reporting database, do the following:
     1. In the pop-up window, next to the `Target database connection` select `Edit...`, and enter the password for the `sqlAdmin` user - this can be found in the `ntbs-ops-dbs-credentials` secret in Azure. (If you also check `Remember password` then you do not need to repeat this step in future).
     2. Click `Generate Script`. Review the SQL script that is generated, and confirm that it is reflective of your changes. In this case, it should be creating all of the tables, and adding all of the functions and stored procedures.
     3. When you are happy with this script, double click on the xml file again and then click `Publish` in the dialog box that pops up. This will generate the same script again, and then run it on the SQL server, creating your new database in the process.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,31 +1,36 @@
...
+3. Publish this database, by double clicking on the xml file in the VS Solution explorer:
+    1. In the pop-up window, next to the `Target database connection` select `Edit...`, and enter the password for the `sqlAdmin` user - this can be found in the `ntbs-ops-dbs-credentials` secret in Azure. (If you also check `Remember password` then you do not need to repeat this step in future).
+    2. Click `Generate Script`. Review the SQL script that is generated, and confirm that it is reflective of your changes. In this case, it should be creating all of the tables, and adding all of the functions and stored procedures.
+    3. When you are happy with this script, double click on the xml file again and then click `Publish` in the dialog box that pops up. This will generate the same script again, and then run it on the SQL server, creating your new database in the process.
+4. To populate the database, first run `uspPopulateCalendarTable`, and then run `uspGenerate`. This database is now ready for use.
```
aptemus, line 22: You also need to run the script which can be found here: https://github.com/publichealthengland/ntbs-reporting/blob/master/source/Scripts/PopulateFeatureFlags.sql  specifying the right values for your environment (for development this will always be 1, 1, 1

aptemus (changes requested):
One minor update to the README @benjimarshall and I suspect I haven't mentioned you need to do that in your development database, sorry about that.

README.md
benjimarshall: I think we did do this, but I then forgot about it. Sorry about that! Added in now

JoshCadney (approved):
Looks good :-)

PR #19, opened 2021-03-18
NTBS-1664 Use TreatmentRegimen in ReusableNotification

benjimarshall:
### Description
* Use `TreatmentRegimen` in the format found in NTBS in the `ReusableNotification` table, instead of the `ShortCourse/MDRTreatment` style as found in ETS. Make the same changes in `ReusableNotification_ETS`.
* Add `uspGenerateReusableNotificationTreatmentRegimen_ETS` to convert ETS-style treatment regimen in NTBS-style, to go into `ReusableNotification_ETS` and then `ReusableNotification`.
* Update `uspMoveRecordsToLegacyExtract` to not get the now-removed `ShortCourse/MDRTreatment` fields from `ReusableNotification`, populate instead with blank placeholders.
* Move the `uspGenerateReusableNotificationTreatmentRegimen` procedure to `uspUpdateLegacyTreatmentRegimen`: it converts NTBS-style `TreatmentRegimen` to ETS-style. Use this to add treatment regimens for NTBS notifications to the `LegacyExtract`.
* Remove `uspReusableNotification` now that the `ReusableNotification` schema has changed, it was used in now unsupported SSRS reports. Similarly, remove references to `ShortCourse/MDRTreatment` in `ufnAuthorizedReusableNotification`, another function used for deprecated SSRS reports.

### Testing
Checked data is being correctly copied from NTBS into the reusable notification, and same for ETS. Checked that LegacyExtract changes have done the expected thing as well, using test data in Azure.


JoshCadney (approved):

source/dbo/Stored Procedures/Reusable Notification/uspGenerateReusableNotification.sql
```diff
@@ -266,34 +265,33 @@ BEGIN TRY
...
 			--treatment details
-			,dbo.ufnYesNo(cd.IsPostMortem)					AS 'PostMortemDiagnosis' 
-			,dbo.ufnYesNo(cd.DidNotStartTreatment)			AS 'DidNotStartTreatment' 
+			,dbo.ufnYesNo(cd.IsPostMortem)					AS 'PostMortemDiagnosis'
+			,dbo.ufnYesNo(cd.DidNotStartTreatment)			AS 'DidNotStartTreatment'
 			--next two fields set in separate function later on
-			,NULL						                    AS 'ShortCourse' 
-			,NULL							                AS 'MdrTreatment' 
-			,cd.MDRTreatmentStartDate						AS 'MdrTreatmentDate' 
+			,cd.TreatmentRegimen							AS 'TreatmentRegimen'
```
aptemus, line 293: This needs to be converted from the 'code' held in clinical details to a user-readable string, e.g. StandardTherapy to Standard therapy, MdrTreatment to 'RR/MDR/XDR treatment'

source/dbo/Stored Procedures/Reusable Notification/uspGenerateReusableNotificationTreatmentRegimen_ETS.sql
```diff
@@ -0,0 +1,43 @@
...
+		Other
+**************************************************************************************************/
+
+
+CREATE PROCEDURE [dbo].[uspGenerateReusableNotificationTreatmentRegimen_ETS]
+
+AS
+
+	UPDATE dbo.ReusableNotification_ETS
+		SET TreatmentRegimen = 'StandardTherapy'
```
aptemus, line 20: As comment above - this needs to output proper phrases, not the 'code' values used in the application.  Unfortunately this enum is held in code only and doesn't live in the ReferenceData schema in NTBS, where we could grab the description from the code. What I've done for other similar issues in the reporting database is create a lookup table which will need to be maintained in sync with the code, e.g. https://github.com/publichealthengland/ntbs-reporting/blob/master/source/dbo/Tables/Reusable%20Notification/OutcomeLookup.sql

aptemus (commented):
@benjimarshall I'm not completely sure why I got a review request now but there are some further updates needed before I can QA.  This problem of shared lookup values is one we encounter every now and then, for such a short list I don't think it's worth revisiting the NTBS implementation of the list purely in code.

PR #21, opened 2021-03-22
NTBS-1664 Use treatment regimen descriptions, not codes

benjimarshall:
## Description

A change missed from #19 review before it was merged into `master`.

Use textual descriptions `('Standard treatment', 'RR/MDR/XDR treatment', 'Other')` instead of codes `('StandardTreatment', 'MdrTreatment', 'Other')` in the new `ReusableNotification.TreatmentRegimen` column.

Add a `TreatmentRegimen` lookup table to convert between the two formats.

## Testing
Published to my Azure dev database, results the same as last time just now encoded correctly. This has not been deployed to any other DB.

### DIFF
--- a/source/dbo/Stored Procedures/Reusable Notification/uspGenerateReusableNotificationTreatmentRegimen_ETS.sql
+++ b/source/dbo/Stored Procedures/Reusable Notification/uspGenerateReusableNotificationTreatmentRegimen_ETS.sql
@@ -16,8 +16,22 @@ CREATE PROCEDURE [dbo].[uspGenerateReusableNotificationTreatmentRegimen_ETS]
 
 AS
 
+	DECLARE @StandardTreatment NVARCHAR(30), @MdrTreatment NVARCHAR(30), @OtherTreatment NVARCHAR(30)
+
+	SELECT @StandardTreatment = TreatmentRegimenDescription
+	FROM TreatmentRegimenLookup
+	WHERE TreatmentRegimenCode = 'StandardTherapy'
+
+	SELECT @MdrTreatment = TreatmentRegimenDescription
+	FROM TreatmentRegimenLookup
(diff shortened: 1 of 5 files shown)


source/dbo/Stored Procedures/Reusable Notification/uspGenerateReusableNotificationTreatmentRegimen_ETS.sql
```diff
@@ -17,15 +17,15 @@ CREATE PROCEDURE [dbo].[uspGenerateReusableNotificationTreatmentRegimen_ETS]
 AS
 
 	UPDATE dbo.ReusableNotification_ETS
-		SET TreatmentRegimen = 'StandardTherapy'
+		SET TreatmentRegimen = 'Standard therapy'
```
aptemus, line 20: I would use the lookup table here as well, just so it is a little bit more maintainable, even if all you do is declare 3 variables at the beginning, one containing one SELECT from the lookup table.

aptemus (changes requested):
@benjimarshall one minor suggestion. Plus you could have used a view, rather than a table, to avoid having to repopulate it in uspSeed. Equally there's nothing wrong with the way you've done it, and it will run faster.

source/dbo/Stored Procedures/Reusable Notification/uspGenerateReusableNotificationTreatmentRegimen_ETS.sql
benjimarshall: Yeah I was mainly following the `OutcomeLookup` table's pattern but I think both view or table would have worked fine. I've gone for the local variable approach as you suggested

JoshCadney (approved):
Looks good :-)

PR #23, opened 2021-03-25
NTBS-1664 Check for nulls when setting treatment regimen

benjimarshall:
### Description
When checking whether an ETS planned treatment is not equal to 1 (eg. `ShortCourseTreatment <> 1`), this condition also needs to return return if the value is `NULL`. Do this using `COALESCE` expressions.

### Testing
Tested on my Azure dev DB, found examples of this issue (eg. `EtsId = 142466`) and checked they no longer are present after the change. I've not deployed this anywhere else.


aptemus (approved):
Good idea, normally I use COALESCE for everything but I was thinking here of an OR IS NULL but this is neater.

PR #24, opened 2021-03-29
NTBS-1664: Fix references to deleted properties in reusable notification

LeoCie:
I was unable to publish a fresh reporting database. Me and Josh found that there are a lot of references to the reusable notification properties in functions/stored procedures which all have to be changed if the reusable notification changes.
We think the issue didn't stop Benji from publishing as the effected areas weren't changed so weren't run again when he re-published.


JoshCadney:
In Benji's previous PR he alluded to the fact that some of these functions/stored procedures are now redundant. So the question is: can we remove any of them entirely? It would certainly give us less to worry about in the upcoming changes to the `ReusableNotification`.

aptemus:
> In Benji's previous PR he alluded to the fact that some of these functions/stored procedures are now redundant. So the question is: can we remove any of them entirely? It would certainly give us less to worry about in the upcoming changes to the `ReusableNotification`.

@JoshCadney - yes, all of the affected SPs are only applicable to SSRS and won't carry through to our Power BI solution.

LeoCie:
This pull request is now redundant as we are removing all now-unused code in NTBS-2200: https://github.com/publichealthengland/ntbs-reporting/pull/28

PR #25, opened 2021-03-30
NTBS-2179: Add columns to reusable notification

LeoCie:
Added Lat, Long and ClusterId to reusable notification. Adding these meant that references to clusterId no longer need to be gotten from joins, so to be able to build this (without ambiguous reference errors) I removed those joins.

### DIFF
--- a/source/dbo/Stored Procedures/Reusable Notification/uspGenerateReusableNotification.sql
+++ b/source/dbo/Stored Procedures/Reusable Notification/uspGenerateReusableNotification.sql
@@ -144,7 +144,10 @@ BEGIN TRY
 			  ,[QUIN]
 			  ,[MDR]
 			  ,[XDR]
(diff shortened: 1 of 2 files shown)


JoshCadney (commented):
It's not quite clear how this relates to your other open PR. If we haven't yet removed the other reports then we need to update them to prevent the same errors you were seeing yesterday. And even if we remove the reports, do we still need the `ReusableNotificationType`?

LeoCie:
> It's not quite clear how this relates to your other open PR. If we haven't yet removed the other reports then we need to update them to prevent the same errors you were seeing yesterday. And even if we remove the reports, do we still need the `ReusableNotificationType`?

All code that would have needed to be changed (the other report code) has now been removed in https://github.com/publichealthengland/ntbs-reporting/pull/28.

JoshCadney (approved):

PR #28, opened 2021-03-31
NTBS-2200: remove code which was used only for SSRS operational reporting

aptemus:
I have removed all of the code which was used only to pull data from the database to display it on the SSRS reports.  There is one exception to this - the SSRS reports in the data migration folder will still remain.  (I also haven't yet removed the RDL files from the `sql_clients` folder as they aren't breaking anything).

What is left is the code which generates the content of reports or performs other functions like creating the `LabSpecimen` table.  I have left in the `uspGenerate[report]` SPs which generate the basic numbers used to calculate the aggregate figures for reports like Culture & Resistance; we will still need something like this for Power BI, albeit probably not quite the same.

I have not yet attempted to reorganise the folders as different groupings probably make sense for the future, but that will come next.

I've tested this in my own database, publishing the code to it brand new to make sure no dependencies got missed.  There is one outstanding question about `[uspNTSReportLabMappings]`, which is referring to an SP `[uspGetAuthenticatedLoginGroups]` which was renamed long ago, so I am assuming this code is probably redundant but belongs to Adil.  We can discuss this with him when he gets back and restore it if need be.

I ran `uspGenerate`, `uspLabSpecimen` and `uspPopulateForestExtract` which should exercise nearly all of the code in the database; all completed without errors.  I have not tested `uspUpdateNotificationClusterMatchWithForestClusterData` as this is called from an SSIS package so only Adil can do it, but no code near it has changed.

I've gone through and tidied up a load of typos which were really annoying me and also not helpful when trying to use the build process to spot warnings.  All the warnings that remain are where we're making calls to the master db for which we don't have a DACPAC.

I commented out `sp_generate_merge` as I have no idea whether we need it any more and it also generates a load of warnings.


benjimarshall (approved):
Looks good!

LeoCie (approved):

PR #29, opened 2021-04-07
NTBS-2125: Add release version table and use in footer

LeoCie:
Now add release version and date to the footer used for reports. 
This PR is linked to https://github.com/publichealthengland/ntbs_Beta/pull/1167


aptemus (commented):
@LeoCie so the bit that's missing is how to actually set the release version.  I think may also be a problem for the application as well, although a different one!

I think for the reporting database, I would have uspSeed populate the Release Version table.  When we want to change the release version, we update uspSeed with the new details.  Does that make sense?  For both the application and the reporting service, we need to decide what our current release version is, so we can start displaying it (maybe something like 'Pre-release', it doesn't have to be a number?)

JoshCadney (approved):
Looks good :-)

PR #30, opened 2021-04-07
NTBS-2188 Improve uspSeed

benjimarshall:
### Changes
* Update README: NTBS-2188 specimen matching changes in publichealthengland/ntbs-specimen-matching#5
* Run `uspPopulateCalendarTable` from `uspSeed` (but only do this during its first run)
* Run code based on `PopulateFeatureFlags.sql` script during initial DB setup from `uspSeed`
* Remove `PopulateFeatureFlags.sql` because it has been inlined into `uspSeed`

This means the entire DB can be set up just by running `uspGenerate`.

### Testing
Tested both acceptance criteria on local DBs: that the calendar table and feature flags are not modified if there already exists records in those tables, and that I can create a new reporting DB and it successfully runs `uspGenerate`, populating the calendar and feature flag tables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,13 +41,11 @@ Steps:
 1. If the instance of SQL Server that you are using is not at `localhost` then update the `Data Source` in the `TargetConnectionString`.
 1. Double-click on this config in the `Solution Explorer` panel in Visual Studio to publish the codebase. This will build the relevant views and set up the relevant tables.
 1. To populate the database, do the following:
-    1. Run the stored procedure `uspSeed` in the specimen matching database.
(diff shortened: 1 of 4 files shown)


aptemus (approved):
@benjimarshall looks good, only question is whether you want to rename uspSeed under this ticket or if we want a separate one to go through and sort this out - it feels like it might be a rather volatile time to do it right now.

benjimarshall:
@aptemus considering that the migration DB has a `uspSeed` that needs renaming, and this ticket doesn't touch that DB, perhaps we could rename that one and the one in this reporting DB in the same new ticket?

PR #31, opened 2021-04-12
Major overhaul of reusable notification structure for Power BI

aptemus:
I have 'replaced' the ReusableNotification table with four tables:
* RecordRegister - the primary table which will have row level security applied to it in Power BI
* Record_PersonalDetails - for those personal details which will be masked in the Cluster Report
* Record_CaseData - for the remaining fields (i.e. most of them) from ReusableNotification
* Record_LegacyExtract - for the fields that only appear in the legacy extract

This code makes some use of the existing code in uspGenerateReusableNotification and so there are places where it is inconsistently applying '' to an empty field rather than NULL.  Following this code review, my plan would be to remove the ReusableNotification table completely and replace it with a view with the same name and content. This gives us backwards compatibility for a period of time whilst we untangle the other dependences on the reporting database, particularly the views which will be moving to the specimen matching database.

The view would have to present mocked-up values for the fields at the end of the table: 
       [CulturePositive]
      ,[Species]
      ,[EarliestSpecimenDate]
      ,[DrugResistanceProfile]
      ,[INH]
      ,[RIF]
      ,[EMB]
      ,[PZA]
      ,[AMINO]
      ,[QUIN]
      ,[MDR]
      ,[XDR]
as these will be calculated in the specimen matching database under NTBS-2227.

I have tested this by comparing the records with the ReusableNotification table in `test`. There are still some very minor discrepancies but by and large the two sources match.  I will attached the SQL statement I've used to compare both to the Jira ticket.

I have tested this against live data and it performs the whole job in under 20 seconds.   At this stage, I don't think an exhaustive comparison of each row of the two tables will be that useful, as the new line list will be slightly different from the old one. It includes new fields which are NTBS-specific.

I have started comparison testing this against the Legacy Extract as well but that is proving more difficult with more discrepancies to chase down.  However, as we have no dependencies on this in the same way, I wanted to submit the code for review before I did any more without any eyes on it.

I haven't asked Adil to be a code reviewer, I've shown him the basic structure of it, but this way he can do QA.


source/dbo/Stored Procedures/Power BI Reporting/uspUpdateRecordPostcode.sql
```diff
@@ -0,0 +1,42 @@
+﻿/*
+	This procedure creates a cleaned postcode from the value supplied from the notification system (ETS or NTBS)
+	Provided the postcode matches the format and length of a UK postcode, this will insert a single space 3 characters from
+	the left hand side of the string
```
JoshCadney, line 4: I think the stored procedure is correct, but the comment should say "from the right hand side of the string"

source/dbo/Stored Procedures/Power BI Reporting/uspPopulateLookupTables.sql
```diff
@@ -0,0 +1,31 @@
...
+AS
+BEGIN TRY
+	TRUNCATE TABLE [dbo].[DOTLookup]
+	INSERT INTO [dbo].[DOTLookup](SystemValue, DOTOffered, DOTReceived)
+	VALUES
+	--ETS values
+	('1', 'Yes', 'DOT received'),
+	('0', NULL, NULL),
+	('2', 'Unknown', 'Unknown'),
+	--NTBS values - IsDotOffered doesn't require translation for the reporting service
```
JoshCadney, line 20: If DotOffered is not needed for reporting, then why do we need to include it for the ETS values?

JoshCadney (changes requested):
I haven't been through every line of this but it generally looks good :-)

Only a couple of minor comments

source/dbo/Stored Procedures/Power BI Reporting/uspPopulateLookupTables.sql
aptemus: @JoshCadney IsDotOffered doesn't require _translation_ for the reporting service; the values are stored as 'Yes', 'No' and 'Unknown' so there's no need to look them up to Yes, No and Unknown.

JoshCadney: Oh, I see - thanks!

JoshCadney (approved):

PR #32, opened 2021-04-14
NTBS-1824: Take account of migrated records and update date on migration reports

LeoCie:
Now only take records from the last 3 full years and also exclude any records that are already migrated into NTBS.

### DIFF
--- a/source/dbo/Stored Procedures/Data Migration/uspConfirmationLineList.sql
+++ b/source/dbo/Stored Procedures/Data Migration/uspConfirmationLineList.sql
@@ -18,10 +18,11 @@ AS
 		LEFT OUTER JOIN  [$(NTBS_R1_Geography_Staging)].[dbo].[TB_Service_to_Hospital] tbh ON tbh.HospitalID = mn.NtbsHospitalId
 		LEFT OUTER JOIN  [$(NTBS_R1_Geography_Staging)].[dbo].[TB_Service_to_PHEC] tbsp ON tbsp.TB_Service_Code = tbh.TB_Service_Code
 		LEFT OUTER JOIN  [$(NTBS_R1_Geography_Staging)].[dbo].[PHEC] p ON p.PHEC_Code = tbsp.PHEC_Code
+		LEFT JOIN vwNotificationYear ny ON ny.NotificationYear = YEAR(mn.NotificationDate)
 	WHERE 
 		mn.GroupId IS NOT NULL
 		AND p.PHEC_Name = @Region
-		AND mn.NotificationDate >= '2017-01-01'),
+		AND ny.Id >= -3),
 	LinkedNotifications AS
 	(SELECT ng.GroupId, STRING_AGG(CAST(mn.PrimaryNotificationId AS NVARCHAR(MAX)), ', ') AS LinkedNotifications
 	FROM [$(migration)].[dbo].[MergedNotifications] mn 
@@ -50,10 +51,15 @@ AS
 		LEFT OUTER JOIN [$(NTBS_R1_Geography_Staging)].[dbo].[PHEC] p ON p.PHEC_Code = tbsp.PHEC_Code
 		LEFT OUTER JOIN NotificationGroups ng ON ng.GroupId = mn.GroupId
 		LEFT OUTER JOIN LinkedNotifications ln ON ln.GroupId = mn.GroupId
+		LEFT JOIN vwNotificationYear ny ON ny.NotificationYear = YEAR(mn.NotificationDate)
 	WHERE 
 		ng.GroupId IS NOT NULL
 	OR 
-		(p.PHEC_Name = @Region AND mn.NotificationDate >= '2017-01-01')
+		(p.PHEC_Name = @Region AND ny.Id >= -3)
+	AND NOT EXISTS
+		(SELECT LegacyId
+		FROM [$(migration)].[dbo].ImportedNotifications impn
+		WHERE LegacyId = EtsId OR LegacyId = LtbrId)
 
 	ORDER BY mn.NotificationDate DESC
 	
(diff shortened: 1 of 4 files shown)


source/dbo/Stored Procedures/Data Migration/uspConfirmationLineList.sql
```diff
@@ -18,10 +18,15 @@ AS
...
 		LEFT OUTER JOIN  [$(NTBS_R1_Geography_Staging)].[dbo].[TB_Service_to_PHEC] tbsp ON tbsp.TB_Service_Code = tbh.TB_Service_Code
 		LEFT OUTER JOIN  [$(NTBS_R1_Geography_Staging)].[dbo].[PHEC] p ON p.PHEC_Code = tbsp.PHEC_Code
+		LEFT JOIN vwNotificationYear ny ON ny.NotificationYear = YEAR(mn.NotificationDate)
 	WHERE 
 		mn.GroupId IS NOT NULL
 		AND p.PHEC_Name = @Region
-		AND mn.NotificationDate >= '2017-01-01'),
+		AND ny.Id >= -3
+		AND NOT EXISTS
+			(SELECT LegacyId
```
aptemus, line 27: Now, I appreciate this SHOULD work but we've never relied on the ImportedNotifications table. I would go to NTBS to see if the value is there, rather than relying on it, but happy to discuss.

source/dbo/Stored Procedures/Data Migration/uspMigrationAdditionalNotificationsByRegionAndYear.sql
```diff
@@ -14,21 +14,31 @@ AS
...
 		LEFT OUTER JOIN  [$(NTBS_R1_Geography_Staging)].[dbo].[TB_Service_to_PHEC] tbsp ON tbsp.TB_Service_Code = tbh.TB_Service_Code
 		LEFT OUTER JOIN  [$(NTBS_R1_Geography_Staging)].[dbo].[PHEC] p ON p.PHEC_Code = tbsp.PHEC_Code
+		LEFT JOIN vwNotificationYear ny ON ny.NotificationYear = YEAR(mn.NotificationDate)
 	WHERE 
 	mn.GroupId IS NOT NULL
 	AND p.PHEC_Name = @Region
-	AND mn.NotificationDate >= '2017-01-01')
+	AND ny.Id >= -3
+	AND NOT EXISTS
+		(SELECT LegacyId
```
aptemus, line 23: Now strictly speaking this shouldn't happen - if one of the group has been migrated into NTBS then they all have.  So we'd probably want to flag separately if we were going to migrate in any records that were somehow grouped with records already in NTBS. I THINK it's too unlikely to worry about.  We would find out during data cleaning anyway and be able to rectify it then.  So I think I would take this check out.

source/dbo/Stored Procedures/Data Migration/uspMigrationAdditionalNotificationsByRegionAndYear.sql
```diff
@@ -14,21 +14,31 @@ AS
...
 		LEFT OUTER JOIN  [$(NTBS_R1_Geography_Staging)].[dbo].[TB_Service_to_Hospital] tbh ON tbh.HospitalID = mn.NtbsHospitalId
 		LEFT OUTER JOIN  [$(NTBS_R1_Geography_Staging)].[dbo].[TB_Service_to_PHEC] tbsp ON tbsp.TB_Service_Code = tbh.TB_Service_Code
 		LEFT OUTER JOIN  [$(NTBS_R1_Geography_Staging)].[dbo].[PHEC] p ON p.PHEC_Code = tbsp.PHEC_Code
+		LEFT JOIN vwNotificationYear ny ON ny.NotificationYear = YEAR(mn.NotificationDate)
 		INNER JOIN NotificationGroups ng ON ng.GroupId = mn.GroupId
 	WHERE mn.GroupId IS NOT NULL
 	AND 
 	(p.PHEC_Name != @Region
-	OR mn.NotificationDate < '2017-01-01')
+	OR ny.Id >= -3)
```
aptemus, line 37: Now, I actually think the existing code is wrong here.  We are going to migrate in any linked record, regardless of whether it's within our time period or not.  So I think we are not displaying them on the report and we should be. I need to look at exactly where this gets called from on the report.

aptemus (changes requested):
@LeoCie A couple of things to discuss and one thing for me to check in the report itself.

JoshCadney (approved):
I don't have anything to add beyond Nancy's comments

source/dbo/Stored Procedures/Data Migration/uspConfirmationLineList.sql
LeoCie: If it's unreliable then changing it seems like the better option. This table is also used in search to filter out already imported notifications, maybe we should look at removing it altogether and just going to the notification table instead?

aptemus: Yes, this is true - either it's reliable or it isn't. And it was put in originally as a safety measure to be sure we didn't offer the user a legacy notification to import that had actually already been imported, this is the ticket: https://airelogic-nis.atlassian.net/browse/NTBS-1440  so on that basis we SHOULD be able to use it.

source/dbo/Stored Procedures/Data Migration/uspMigrationAdditionalNotificationsByRegionAndYear.sql
LeoCie: I've taken this out for now, if we decide it should be there then we can put it back

source/dbo/Stored Procedures/Data Migration/uspConfirmationLineList.sql
LeoCie: I have commit a change that uses the NTBS notifications table instead. It seems like it's the safest option, as the app showing a legacy notification seems a little less risky than a reporting showing false data. I did the change in one commit, so if we decide to use the imported table then reverting shouldn't be too bad.

source/dbo/Stored Procedures/Data Migration/uspMigrationAdditionalNotificationsByRegionAndYear.sql
LeoCie: I have readded this as we realised that it's needed to get the additional notifications. I have also updated the comment

PR #33, opened 2021-04-15
NTBS 2217 delete profiles

aptemus:
This PR is mainly for the output of our investigation into how to conditionally drop objects from the target database scheme when publishing changes.  Most of the time we don't want to do this, and have our publish profiles set not to do so. So the simplest way to ensure it can be done was to create a parallel publish profile for each environment with different settings.

To aid navigation, I then organised all the publish profiles into folders, and I added a README to the new 'DELETE' folder to explain what its purpose is.

We discovered a problem with the DACPACs which had been checked in recently, so corrected those.

I tested this by running the DELETE profile against a new database on Azure with some dummy objects I wanted to have removed, and some users I did not.   The objects were removed but the users were preserved.  I then also run this against `phe-uat`, and this has successfully removed the unused objects from that database, without removing the users (which in any case I don't have permissions to do).


JoshCadney (approved):
Looks good! :-)

PR #34, opened 2021-04-15
NTBS-2217: remove redundant tables

aptemus:
Removed `ReusableNotification` and `ReusableNotification_ETS` tables - the DELETE publish profiles will need to be used. I've tested this on one of the branch databases on Azure.


JoshCadney (approved):
Looks good :-)

PR #35, opened 2021-04-20
NTBS-2254: streamline user permissions view so it only contains user …

aptemus:
…principal name and code

These are the only values needed for Power BI row-level security


LeoCie (approved):
Looks good to me

benjimarshall (approved):
Looks good!

PR #36, opened 2021-04-21
NTBS-2254: updated DACPAC to take account of tickets in the 'move to …

aptemus:
…specimen matching' Epic, added view for new fields for Power BI and updated ReusableNotification to pull the culture and resistance fields from the new view


benjimarshall (approved):
Looks good!

PR #37, opened 2021-04-28
NTBS-2261: Update forest extract population to use NTBS notification as base, not reusable notification

LeoCie:


source/dbo/Stored Procedures/Extracts/uspPopulateForestExtract.sql
```diff
@@ -69,71 +69,85 @@ BEGIN
...
 			'Positive'																						AS SmearSampleResult,
 			lab.SpecimenDate																				AS SpecimenDate,
 			lab.ReferenceLaboratoryNumber																	AS ReferenceLaboratoryNumber,
 			GETUTCDATE()																					AS ExtractDate,
 			patient.LocalPatientId																			AS LocalPatientId,
-			Age																								AS Age,
-			--hospital.CaseManagerUsername																	AS OwnerUserId,
-			CaseManager																						AS CaseManager,
-			rn.Consultant																					AS PatientsConsultant,
+			(0 + Convert(Char(8),GETDATE(),112) - Convert(Char(8),patient.Dob,112)) / 10000					AS Age,
```
JoshCadney, line 120: What is happening on this line?

source/dbo/Stored Procedures/Extracts/uspPopulateForestExtract.sql
```diff
@@ -69,71 +69,85 @@ BEGIN
...
+			--hospitalDetails.CaseManagerUsername															AS OwnerUserId,
```
JoshCadney, line 121: I think this comment can be removed

JoshCadney (commented):
Looks good - but one question

source/dbo/Stored Procedures/Extracts/uspPopulateForestExtract.sql
aptemus: I think that's a very complicated way to work out the patient's age at time of notification - we have a function that does that, which can be called by passing in date of birth and notification date - https://github.com/publichealthengland/ntbs-reporting/blob/master/source/dbo/Functions/Reusable%20Notification/ufnGetAgefrom.sql

LeoCie: Using the function now, I thought we might have one but forgot to look

JoshCadney (approved):
Looks good :-)

PR #41, opened 2021-05-04
NTBS-2231 Remove unused specimen matching features

benjimarshall:
### Description
Remove various Tables, Views, Stored Procedures and Functions that were used when doing parts of the specimen matching process in reporting, but that has all moved over as part of the NTBS-2235 epic so we can remove all these unused bits now.

### Testing
I have successfully deployed all of this locally and `uspGenerate` runs and produces the expected result as far as I can tell. KPIs still can be retrieved by the NTBS app, which is the only Reporting -> NTBS dependency left. I have not been able to check whether this has broken PowerBI or not but I'm fairly sure it shouldn't.

### Outstanding questions
* There were a few SPs and Views left after removing all of the items listed in the ticket, and I was wondering whether these can be removed as well or not? In particular, the contents of `Stored Procedures/Lab Results`, `Views/Lab Results`, and `Views/Reusable Drop Downs/vwSpecies.sql`
* There were a few references to a few of the SPs and tables I have removed in a few of the SSRS reports. I haven't found a way of modifying them, and I suspect many of them will be being removed soon anyway. What should I do about this?


JoshCadney (approved):
Looks good to me :-)

This is a very satisfying PR!

benjimarshall:
Based on a conversation with Nancy the only SSRS reports left that we care about are the reports in `sql-clients\ssrs\Data Migration`. We have broken one of these reports (`Data Migration/SOP Specimen Match Multi System.rdl`) with this due to now missing references to `LabSpecimen` and `EtsSpecimenMatch`. These need to be retrieved from the specimen-matching DB, but I don't think I am able to edit and test this properly, so I will raise a follow on ticket

PR #42, opened 2021-05-04
NTBS-2259: Add new stored procedure for chest xray results

LeoCie:
We are "ranking" the chest x-ray results: firstly based on how much time there is between result and diagnosis and then on the result itself. We take the top ranked result and add this to the record.


JoshCadney (approved):
Looks good to me :-)

source/dbo/Stored Procedures/Power BI Reporting/uspMapCaseRecordChestXrayResults.sql
```diff
@@ -0,0 +1,36 @@
...
+			WHEN Result = 'Awaiting' THEN 4 END ASC
+		) AS rn
+	FROM [$(NTBS)].[dbo].[ManualTestResult] mr
+		INNER JOIN [dbo].[Record_CaseData] cd ON cd.NotificationId = mr.NotificationId
+	WHERE ManualTestTypeId = 4
+	)
+	UPDATE cd
+	SET ChestXRayResult =
+	CASE
+	WHEN rankedResults.result = 'Awaiting' THEN 'Awaiting'
```
aptemus, line 23: I would probably have put this in a lookup table as we have with other values we have to decode from NTBS, but I don't think this has to be done. It does mean the ranking could be stored in the same table, which would eliminate both of the CASE statements. I feel like this would be more performant at higher volumes?

source/dbo/Stored Procedures/Power BI Reporting/uspMapCaseRecordChestXrayResults.sql
```diff
@@ -0,0 +1,36 @@
...
+	,ROW_NUMBER() OVER
+		(PARTITION BY mr.NotificationId
+		ORDER BY ABS(DATEDIFF(day, mr.TestDate, cd.DiagnosisDate)) ASC
+		,CASE WHEN Result = 'ConsistentWithTbCavities' THEN 1
+			WHEN Result = 'ConsistentWithTbOther' THEN 2
+			WHEN Result = 'NotConsistentWithTb' THEN 3
+			WHEN Result = 'Awaiting' THEN 4 END ASC
+		) AS rn
+	FROM [$(NTBS)].[dbo].[ManualTestResult] mr
+		INNER JOIN [dbo].[Record_CaseData] cd ON cd.NotificationId = mr.NotificationId
```
aptemus, line 17: I would add an extra clause here and SourceSystem = 'NTBS' - just belt and braces as Record_CaseData contains ETS records as well. The inner join should never connect ETS data to an NTBS record but it is just a failsafe.

aptemus (commented):
@LeoCie looks good - I _think_ a lookup table might simplify but don't know for sure!

source/dbo/Stored Procedures/Power BI Reporting/uspMapCaseRecordChestXrayResults.sql
JoshCadney: Yes, I think this would be a good idea, and probably more performant

source/dbo/Stored Procedures/Power BI Reporting/uspMapCaseRecordChestXrayResults.sql
LeoCie: Record_CaseData doesn't have a source field, it's RecordRegister that does. It would make sense to have this check, maybe we should add in the source into Record_CaseData?

PR #43, opened 2021-05-07
NTBS-2231 Add known specimen matching issues to README

benjimarshall:
QA addition: Document known issues with running `uspGenerate` in specimen-matching, related to having small amounts of data.


aptemus (approved):
👍

PR #44, opened 2021-05-11
NTBS-2261: Fix fields in population of forest extract

LeoCie:
Some fixes that came back from QA. Mainly just mapping 1/0's to Yes/No or Yes/blank.
Also I'd joined the residence phec on the wrong field, which has now been fixed and is populating properly.


JoshCadney (approved):
Looks good :-)

PR #46, opened 2021-05-12
NTBS-2239 and NTBS-2290

aptemus:
Made updates to the migration reports so that:
* they can run against the EtsSpecimenMatch table in the specimen matching database, instead of in the reporting database
* they are properly parameterised, so we can deploy these against the live and UAT instances as well as dev (this partly involved moving queries out of SSRS and into stored procedures, which is why the code appears new)
* the reporting period is finally set correctly back to current year + 3 previous, not +4 previous, which is a hangover to a long-ago change made for the East of England region

I have also removed some dependencies on a table called 'TB Service short names' which I used to populate the sheet names on a report which was intended for extract to Excel.  The table has gone; I must never have committed it to source code and then a recent delete publish profile has removed it from `phe-dev`. So if it is required again, I will recreate it.


JoshCadney (approved):
Looks good to me :-)

PR #47, opened 2021-05-13
NTBS-2259: Add default value for no chest xray result

LeoCie:
Added a default value for no chest x-ray result.


aptemus (approved):
@LeoCie 👍

PR #48, opened 2021-05-13
NTBS-2273: Add new column with formatted nhs number to record patient details

LeoCie:
We want a formatted NHS number in both columns:
- In the NhsNumber column we want 000 000 0000 
- In the NhsNumberToLookup column we want 0000000000

Unless the NHS number has alphabetic characters, in which case we want NhsNumber to be unchanged.

### DIFF
--- a/source/dbo/Stored Procedures/Power BI Reporting/uspUpdateRecordNhsNumber.sql
+++ b/source/dbo/Stored Procedures/Power BI Reporting/uspUpdateRecordNhsNumber.sql
@@ -1,0 +1,24 @@
+﻿CREATE PROCEDURE [dbo].[uspUpdateRecordNhsNumber] AS
+	SET NOCOUNT ON
+
+	BEGIN TRY
+
+		UPDATE [dbo].[Record_PersonalDetails]
+		SET NhsNumberToLookup =
+			CASE
+				WHEN LEN(REPLACE(REPLACE(NhsNumberToLookup, ' ', ''), '-', '')) = 10
+					THEN REPLACE(REPLACE(NhsNumberToLookup, ' ', ''), '-', '')
+				ELSE NULL
+			END
+
+		UPDATE [dbo].[Record_PersonalDetails]
+			SET NhsNumber = SUBSTRING(NhsNumberToLookup, 1, 3) +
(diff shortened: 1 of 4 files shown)


source/dbo/Stored Procedures/Power BI Reporting/uspUpdateRecordNhsNumber.sql
```diff
@@ -0,0 +1,16 @@
+﻿CREATE PROCEDURE [dbo].[uspUpdateRecordNhsNumber] AS
+	SET NOCOUNT ON
+
+	BEGIN TRY
+
+		UPDATE [dbo].[Record_PersonalDetails]
+			SET NhsNumber = SUBSTRING(NhsNumberToLookup, 1, 3) +
+				' ' + SUBSTRING(NhsNumberToLookup, 4, 3) +
+				' ' + SUBSTRING(NhsNumberToLookup, 7, 4)
+			WHERE NhsNumberToLookup NOT LIKE '%[^0-9]%'
```
aptemus, line 10: @LeoCie I think you've got two NOTs in this patindex.  The ^ means 'match any character not between the square brackets' so we're saying any NHS number which isn't not like 0 .. 9 !

We also want to make sure the string is 10 characters long before we begin, no point formatting if it isn't a valid NHS number to start with.

aptemus (changes requested):
@LeoCie just one thing to look at I think

source/dbo/Stored Procedures/Power BI Reporting/uspUpdateRecordNhsNumber.sql
LeoCie: Yes the double not is intentional. If I do `LIKE '%[0-9]%'` it would let through things such as `abc9102030abc`. 
I will change it so we check the length.

LeoCie: I have now added in length verification

JoshCadney: It's frustrating that there isn't a more intuitive way to do this... but I don't see one

JoshCadney (approved):
Looks good to me

PR #49, opened 2021-05-13
NTBS-2275 Map ETS birth countries to NTBS in Record_CaseData

benjimarshall:
### Description
Extract ETS countries of birth using its ID and also whether the patient is marked as `UkBorn`, and map the extracted country code to an NTBS-style mixed-case name.

### Testing
Tested using some ETS records I manipulated locally. This has not been deployed anywhere


aptemus (approved):
@benjimarshall 👍

PR #50, opened 2021-05-13
NTBS-2321: conversion issues uncovered in reporting code

aptemus:
Unclear really why these have appeared now but after newly migrating the data from South West, North East and Wales, and then running uspGenerate in the reporting database, a number of new errors have appeared

### DIFF
--- a/source/dbo/Stored Procedures/Power BI Reporting/uspGenerateReportingLegacyExtract.sql
+++ b/source/dbo/Stored Procedures/Power BI Reporting/uspGenerateReportingLegacyExtract.sql
@@ -212,7 +212,7 @@ BEGIN TRY
           ,[ShortCourse]
           ,[InPatient]
           ,[Comments]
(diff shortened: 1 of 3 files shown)


source/dbo/Stored Procedures/Power BI Reporting/uspGenerateReportingLegacyExtract.sql
```diff
@@ -212,7 +212,7 @@ BEGIN TRY
           ,[ShortCourse]
           ,[InPatient]
           ,[Comments]
-          ,[OtherComments]
+          ,LEFT([OtherComments], 60) AS ImmunosuppressionComments
```
JoshCadney, line 215: Why are we using 60 here, but 100 below?

aptemus: Yes, I'm not sure there is a reason - will update above to 100.

JoshCadney (approved):
Looks good :-)

PR #52, opened 2021-05-17
NTBS-2317 Fix a treatment event mapping in migration report

benjimarshall:
### Description
When a patient has not completed a full course and there is no further
reason, the summary should be "Not evaluated" rather than "Lost to
follow-up".

### Testing
I've not managed to test this locally but given how straightforward the change is I'm pretty confident it will work. It builds and deploys locally ok


JoshCadney (approved):
Looks good :-)

PR #54, opened 2021-05-25
NTBS-2350 Set version number in post-deploy script

JoshCadney:
Set version number in post-deploy script rather than during seeding

I have also removed some files which were not doing anything


LeoCie (approved):
Looks good!

PR #55, opened 2021-05-26
NTBS-2351 Rename phe-dev NTBS DB

benjimarshall:
Rename the NTBS DB used by `phe-dev`, 'NTBS_Live' to 'NTBS_Dev'

This change has already been deployed


JoshCadney (approved):
Looks good :-)

PR #56, opened 2021-05-27
NTBS-2356: Add reference data seeding info to readme

LeoCie:


JoshCadney (changes requested):
As mentioned in the PR in the web app project, I don't think we should make these changes here, so I think you can delete this branch/PR.

PR #58, opened 2021-05-27
NTBS-2358 Use PHE UAT migration DB instead of live

benjimarshall:
Use the PHE UAT migration DB that has recently been created instead of the live one for UAT reporting,

This has not been deployed anywhere


JoshCadney (approved):
Looks good :-)

PR #59, opened 2021-05-27
NTBS 2290

aptemus:
I have redone part of this as I didn't realise I had created the branch but never actually created a pull request.


JoshCadney (approved):
I haven't been through every detail, but this looks much cleaner than what we had before :-)

PR #61, opened 2021-06-04
NTBS-2355 Fix records without X-rays getting ignored

benjimarshall:
### Description
Fix records without chest x-rays getting missed by the Record_CaseData
and ForestExtract generating procedures.

This was an issue because ufnGetCaseRecordChestXrayResults returns
nothing if the notification doesn't have a chest X-Ray, so the
notification gets discarded by a cross apply, but not an outer apply.

### Testing
Tested (more thoroughly this time!) on local data. This has not been deployed anywhere


JoshCadney (approved):
Looks good :-)

PR #63, opened 2021-06-08
NTBS-2389 Keep Record_PersonalDetails.Postcode when invalid

benjimarshall:
## Description
When populating `Record_PersonalDetails.Postcode`, first set this to the NTBS `Patients.Postcode`. This is what the user entered, and may be invalid, and may not be formatted correctly.

Replace `Record_PersonalDetails.Postcode` with a formatted version of `Patients.PostcodeToLookup`, if it exists and is valid. If it doesn't exist or is invalid, leave `Record_PersonalDetails.Postcode` as what the user originally entered.

## Testing
Tested locally with an amended version of ETS to have BPFO postcodes. This has not been deployed anywhere

### DIFF
--- a/source/dbo/Stored Procedures/Power BI Reporting/uspUpdateRecordPostcode.sql
+++ b/source/dbo/Stored Procedures/Power BI Reporting/uspUpdateRecordPostcode.sql
@@ -1,42 +1,42 @@
 ﻿/*
 	This procedure creates a cleaned postcode from the value supplied from the notification system (ETS or NTBS)
 	Provided the postcode matches the format and length of a UK postcode, this will insert a single space 3 characters from
-	the right hand side of the string
+	the right hand side of the string. Otherwise, it will leave the postcode value unchanged
 */
 
 CREATE PROCEDURE [dbo].[uspUpdateRecordPostcode] AS
 	SET NOCOUNT ON
 
 	BEGIN TRY
-		
 
-		IF object_id('tempdb.dbo.#RecordPostcodes','U') IS NOT NULL 
-		BEGIN	
+
+		IF object_id('tempdb.dbo.#RecordPostcodes','U') IS NOT NULL
+		BEGIN
 			DROP TABLE #RecordPostcodes
 		END
 
-		SELECT DISTINCT 
+		SELECT DISTINCT
 			PersonalDetailsId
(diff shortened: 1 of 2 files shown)


source/dbo/Stored Procedures/Power BI Reporting/uspUpdateRecordPostcode.sql
```diff
@@ -1,42 +1,44 @@
...
+
+		IF object_id('tempdb.dbo.#RecordPostcodes','U') IS NOT NULL
+		BEGIN
 			DROP TABLE #RecordPostcodes
 		END
 
-		SELECT DISTINCT 
+		SELECT DISTINCT
 			PersonalDetailsId
 			,REPLACE(PostcodeToLookup, ' ', '')	AS CleanedPostcode
```
JoshCadney, line 20: Is this replace necessary? Isn't `PostcodeToLookup` already in the standardised form without spaces (or null)?

source/dbo/Stored Procedures/Power BI Reporting/uspUpdateRecordPostcode.sql
```diff
@@ -1,42 +1,44 @@
...
 		INTO #RecordPostcodes
 		FROM [dbo].[Record_PersonalDetails]
-  
-		DELETE from #RecordPostcodes 
-		WHERE LEN(CleanedPostcode)<5 OR LEN(CleanedPostcode) > 7 OR CleanedPostcode NOT LIKE '%[^0-9]%'
+
+		DELETE from #RecordPostcodes
+		WHERE CleanedPostcode IS NULL
+			OR LEN(CleanedPostcode)<5
+			OR LEN(CleanedPostcode) > 7
```
JoshCadney, line 27: The inconsistency in spacing here is hard to look at!

benjimarshall: Yep, fixed now!

source/dbo/Stored Procedures/Power BI Reporting/uspUpdateRecordPostcode.sql
benjimarshall: Not strictly, no I don't think. This code was already here before but I think it should be safe to remove, so I've done that and refactored a little.

JoshCadney (approved):
Looks good :-)

PR #64, opened 2021-06-08
NTBS-2397 Use specific HIV test values in reports

benjimarshall:
## Description
* Use the more specific 'Offered' values (for 'and done', 'but not done' and 'but refused') in `Record_CaseData`. Use the old `ufnGetHivValue` stored proc to map from the NTBS values to these (basically adding spaces). 
* ETS has the values as we want them, so remove the mapping.
* Remove the now unneeded mapping functions `ufnGetHivTestOffered` and `ufnGetHivTestOffered_ETS`
* Update the KPI views to account for the different types of 'Offered' values.

## Testing
Tested locally with some data in my local NTBS database. This has not been deployed anywhere.


JoshCadney (approved):
Looks good :-)

PR #66, opened 2021-06-09
NTBS-2403: Calculate UK born in reporting database

LeoCie:
Calculate NTBS UK born values and bring them in line with ETS.

PR in app repo: https://github.com/publichealthengland/ntbs_Beta/pull/1312


benjimarshall (approved):
Looks good!

source/dbo/Functions/Power BI Reporting/ufnUkBorn.sql
```diff
@@ -0,0 +1,15 @@
...
+	@CountryId int
+)
+	RETURNS [nvarchar](10)
+AS
+	BEGIN
+		DECLARE @IsUkBorn nvarchar(10)
+		SET @IsUkBorn =
+		CASE 
+			WHEN @CountryId = 235 THEN 'Yes'
+			WHEN @CountryId = 238 THEN 'Unknown'
```
JoshCadney, line 11: Can the `CountryId` passed in here ever be NULL? If it can, would it be better to map NULL to Unknown?

aptemus: Country of birth is a mandatory field, whereas 'Unknown' is an actual value you can select through the front end.  So I think we would want that to output a null if it did happen, to make it easier to spot.

LeoCie: Change made to output null if country id is null

PR #68, opened 2021-06-10
Ntbs 2329 - use ETS as source for Legacy Extract

aptemus:
NTBS-2392 (not 2329 as indicated in the branch name)

Following the review of differences between ETS and NTBS legacy extracts, we've reverted to using the ETS dataexport table as the source for ETS records. These can be presented as a view in SQL Server and then merged into the dataset in Power BI.

One field (ResolvedResidenceHPU) is used in another line list, so that has been 'promoted' to Record_CaseData so it can be reused.

The field names in the view match the names of the fields in Power BI at the point in the ETL where the data is merged.  I've tested this with Power BI to ensure it merges correctly.


JoshCadney (commented):
One very minor thing is that there's a typo in the ticket number on the review/branch - it should be 2392 instead of 2329.

I'd like to understand the motivation for this change a bit better - why is it better to merge the data in Power BI rather than in the reporting database?

JoshCadney (approved):

PR #67, opened 2021-06-09
NTBS-2414 Add other disease site text to legacy extract

benjimarshall:
### Description
Add the `SiteDescription` text stored for 'other' disease sites in NTBS to the legacy extract.

### Testing
Tested locally using ETS test data. This change has not been deployed anywhere yet.

PR #69, opened 2021-06-10
NTBS-2414 Minor refactor of legacy disease site mapping

benjimarshall:
Just a minor review markup from #67 refactoring `uspGenerateReportingLegacySitesOfDisease`


JoshCadney (approved):
This looks good, although it looks like you've somehow ended up with some changes from another branch as well, so you may need to be careful when merging.

benjimarshall:
Ah thanks for catching that! All sorted before I merged

PR #71, opened 2021-06-10
NTBS-2432: No Sample taken

aptemus:
In our general reporting world, we have a field called 'Sample Taken' rather than the value which has always been used historically, 'No Sample Taken'. For the Legacy Extract, we thus need to create a reverse value.

We also noted during testing that if World Region is empty, the Legacy Extract outputs 'Unknown' in ETS so have reproduced that as well.


benjimarshall (approved):
Looks good!

JoshCadney (approved):
Looks good :-)

PR #72, opened 2021-06-10
NTBS-2426 consistent publish profiles

JoshCadney:
Structure the publish profiles in a consistent way across all three database projects.
Clean up compiler warnings.


aptemus (approved):
Looks good, but tedious to do!

JoshCadney:
> Looks good, but tedious to do!

Yes - what's most worrying is that if we want to make a change to one of the settings in our publish profiles, we have to make that change in ~25 different files across 3 repos. I looked into whether there was a better solution, and the answer was "yes, but it requires a fair amount of set up work". So for now I think we should wait and see how often we do want to make changes (I suspect it won't be very often).

PR #73, opened 2021-06-11
NTBS-2426 Fix PHE DEV deploy script

benjimarshall:
The wrong database was being referenced so it wasn't working. I've used this script just now and it now works again.

As an aside, why is there a deploy script for `phe-live` in both the `Pre-production` and `Production` folders?


JoshCadney:
> The wrong database was being referenced so it wasn't working. I've used this script just now and it now works again.
> 
> As an aside, why is there a deploy script for `phe-live` in both the `Pre-production` and `Production` folders?

I think that must have been an oversight - please could you remove the one from `Pre-production` as part of this?

source/Publish Profiles/Pre-production/phe-dev-reporting.publish.xml
```diff
@@ -91,7 +91,7 @@
       <Value>NTBS_Migration</Value>
     </SqlCmdVariable>
     <SqlCmdVariable Include="NTBS">
-      <Value>NTBS_Live</Value>
+      <Value>NTBS_Dev</Value>
```
JoshCadney, line 94: How odd - this looks like it's been this way for a while

JoshCadney: Oh, so I think this has been wrong ever since the delete profiles were created, but has only come to the surface now that we've got rid of the non-delete profiles.

JoshCadney (changes requested):
Looks good :-)

Let's remove the unnecessary profile too

JoshCadney (approved):
Looks good :-)

PR #74, opened 2021-06-16
NTBS-2419 Add Azure UAT migration DB

benjimarshall:
Add new UAT migration DB to deploy script.

As with the change to migration, what branch should I be looking to merge this into? Maybe `master` to then merge into `future-work`?


JoshCadney (approved):
Looks good :-) As with the other one, I can't think of a reason why we'd want to deploy to UAT reporting from any branch other than `future-work` so it's probably ok to merge straight to `future-work`. There's not really a downside to merging to `master` first, it's just more effort!

PR #75, opened 2021-06-16
NTBS-2452: remove erroneous reference to 'EtsLabbase' 

aptemus:
Removed from publish profiles; a warning about this unreferenced SQLCMD variable appears during the release process.


adilmirzaphe (approved):
Looks good

JoshCadney (approved):
Looks good, however the merge may be difficult because this is branched off `master` before I renamed the publish profile changes. It may be easiest to just make this change again on an up to date master branch.

PR #77, opened 2021-06-21
NTBS-2469 - updated text of footer

aptemus:
This is to make the text read better when the Power BI version is added to it during the load to Power BI.  When displayed, the text should now read: 

`Reporting database version: Release-<number>-<date> Power BI Version: Release-<number>-<date>`


JoshCadney (approved):
Looks good :-)

PR #78, opened 2021-06-22
NTBS-2422 Add TableCounts table and stored proc

benjimarshall:
## Description
Add a table to keep track of the important migration and specimen matching tables and views, counting the number of records in them so the app can warn if these tables have become shorter.

## Testing
Tested locally, with [the NTBS app changes for this ticket](https://github.com/publichealthengland/ntbs_Beta/pull/1334). This has not been deployed anywhere yet.


source/dbo/Stored Procedures/Monitoring/uspUpdateTableCounts.sql
```diff
@@ -0,0 +1,77 @@
...
+		,ETS_NotificationCount
+		,LTBR_DiseasePeriodCount
+		,LTBR_PatientEpisodeCount
+		,NotificationClusterMatchCount
+		,NotificationSpecimenMatchCount
+		,EtsSpecimenMatchCount
+	)
+
+	SELECT
+		GETDATE() AS CountTime
```
aptemus, line 24: Could we use GETUTCDATE() here just so we know for sure what time it ran (not very important but I do tend to use GETUTCDATE() whenever possible)

source/dbo/Stored Procedures/Monitoring/uspUpdateTableCounts.sql
```diff
@@ -0,0 +1,77 @@
+﻿CREATE PROCEDURE [dbo].[uspUpdateTableCounts]
+AS
+	INSERT INTO [TableCounts] (CountTime
```
aptemus, line 3: Should this check there isn't already a row for today's date? I'm not sure this would add very much value.

aptemus (commented):
@benjimarshall have put in two very minor comments, both of which I think can be ignored if any of the other reviewers disagree.

source/dbo/Stored Procedures/Monitoring/uspUpdateTableCounts.sql
benjimarshall: So they both give the time to the nearest millisecond, but `GETUTCDATE` does it in UTC and `GETDATE` in the local timezone. I agree using `GETUTCDATE` is a better idea, and I've switched that over

source/dbo/Stored Procedures/Monitoring/uspUpdateTableCounts.sql
benjimarshall: Yeah I suspect this will add more complications than it is worth. Also, we might want to run the job after a release and check the counts haven't gone down, even if the job ran earlier that morning. I think I'll pass on this one, unless anyone objects?

JoshCadney: Agreed, I don't think we want to limit how often the job can run

JoshCadney (approved):
Looks good :-)

PR #79, opened 2021-06-25
NTBS-2488: added view which counts number of notifications by month a…

aptemus:
…nd cluster

I've done this in SQL server rather than Power BI partly for performance and partly because the code is easier to write (and the data would be used to graph epi curves regardless of the reporting solution).


JoshCadney (approved):
Looks good :-)

PR #80, opened 2021-06-30
NTBS-2502 initial code

aptemus:
This is an initial version of the NOIDS extract view.  I have created it as a view because it will effectively be materialised when it is pushed to Power BI.  

The rules for the Specificity_Code field still need to be validated with the national team.
There is also some substitution of Local Authority names which may require another mapping statement in the CTE, to be confirmed tomorrow.


JoshCadney (commented):
Looks good so far :-)

JoshCadney (approved):

adilmirzaphe (commented):
Nice!

PR #81, opened 2021-07-01
NTBS-2500 Add unmatched duplicate report view

benjimarshall:
### Description
Add a report for use in PowerBi for basic duplicate identification across NTBS and ETS, for records with identical NHS numbers and dates of birth

### Testing
View query tested on PHE dev using live data. It takes a couple of seconds to run (but will be materialised in a PowerBi report table eventually so this is no issue)

### DIFF
--- a/source/dbo/Views/Power BI Reporting/vwUnmatchedDuplicates.sql
+++ b/source/dbo/Views/Power BI Reporting/vwUnmatchedDuplicates.sql
@@ -1,0 +1,128 @@
+﻿CREATE VIEW [dbo].[vwUnmatchedDuplicates]
+AS
+
+-- NTBS Notifications, with all notifications migrated in from ETS
+WITH NtbsNotifications AS (
+	SELECT n.NotificationId AS NtbsId,
+		CAST(n.ETSID AS INT) AS EtsId,
+		'NTBS' AS SourceSystem,
+		n.GroupId AS NtbsGroupId,
+		etsGroupings.GroupId AS EtsGroupId,
+		n.NotificationDate,
+		n.CreationDate,
+		COALESCE(cd.TreatmentStartDate, cd.DiagnosisDate, n.NotificationDate) AS TreatmentStartDate,
+		p.Dob,
+		CONCAT_WS(', ', UPPER(p.FamilyName), p.GivenName) AS FullName,
+		COALESCE(tbs.[Name], h.[Name]) AS TbServiceName,
+		phec.[Name] AS PhecName,
+		REPLACE(p.NhsNumber, ' ', '') AS NhsNumber,
+		COALESCE(u.EmailPrimary, u.Username) AS CaseManager
+	FROM [$(NTBS)].dbo.[Notification] n
+		INNER JOIN [$(NTBS)].dbo.Patients p ON n.NotificationId = p.NotificationId
+		INNER JOIN [$(NTBS)].dbo.ClinicalDetails cd ON n.NotificationId = cd.NotificationId
+		INNER JOIN [$(NTBS)].dbo.HospitalDetails hd ON n.NotificationId = hd.NotificationId
+		LEFT JOIN [$(NTBS)].ReferenceData.Hospital h ON hd.HospitalId = h.HospitalId
+		LEFT JOIN [$(NTBS)].ReferenceData.TbService tbs ON h.TBServiceCode = tbs.Code
+		LEFT JOIN [$(NTBS)].ReferenceData.PHEC ON tbs.PHECCode = PHEC.Code
+		LEFT JOIN [$(NTBS)].dbo.[User] u ON hd.CaseManagerId = u.Id
+		LEFT JOIN [$(migration)].dbo.EtsGroupings etsGroupings ON n.ETSID = etsGroupings.EtsId
+	WHERE n.NotificationStatus IN ('Notified', 'Closed')
+),
+NtbsLastRecordTreatmentOutcomeDates AS (
+	SELECT n.NtbsId,
+		MAX(te.EventDate) AS LastRecordedTreatmentOutcomeDate
+	FROM [$(NTBS)].dbo.TreatmentEvent te
+		INNER JOIN NtbsNotifications n ON te.NotificationId = n.NtbsId
+	WHERE te.TreatmentEventType = 'TreatmentOutcome'
+	GROUP BY n.NtbsId
+),
+-- ETS notifications that have not been migrated into NTBS
+EtsNotifications AS (
+	SELECT NULL AS NtbsId,
+		n.LegacyId AS EtsId,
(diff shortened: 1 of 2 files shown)


source/dbo/Views/Power BI Reporting/vwUnmatchedDuplicates.sql
```diff
@@ -0,0 +1,82 @@
...
+	SELECT NULL AS NtbsId,
+		n.LegacyId AS EtsId,
+		NULL AS NtbsGroupId,
+		etsGroupings.GroupId AS EtsGroupId,
+		n.NotificationDate,
+		n.AuditCreate AS CreationDate,
+		p.DateOfBirth,
+		p.Surname AS FamilyName,
+		REPLACE(p.NhsNumber, ' ', '') AS NhsNumber,
+		su.Username AS CaseManager
```
aptemus, line 34: The su.Email may be a bit easier to read, as some of the users have usernames which are like <initial><surname><something that identifies the TB service> e.g. npickeringnewcastle.

source/dbo/Views/Power BI Reporting/vwUnmatchedDuplicates.sql
```diff
@@ -0,0 +1,82 @@
...
+	FROM [$(ETS)].dbo.[Notification] n
+		INNER JOIN [$(ETS)].dbo.Patient p ON n.PatientId = p.Id
+		INNER JOIN [$(ETS)].dbo.SystemUser su ON n.OwnerUserId = su.Id
+		LEFT JOIN [$(NTBS)].dbo.[Notification] ntbsNotification ON n.LegacyId = ntbsNotification.ETSID
+		LEFT JOIN [$(migration)].dbo.EtsGroupings etsGroupings ON n.LegacyId = etsGroupings.EtsId
+	WHERE n.AuditDelete IS NULL AND n.Submitted = 1 AND n.DenotificationId IS NULL
+		AND ntbsNotification.NotificationId IS NULL -- ie no corresponding NTBS migrated notification
+),
+AllNotifications AS (
+	SELECT *
```
aptemus, line 44: Worth including a SourceSystem here for debugging purposes?  It would also prevent two identical records from the two different systems being collapsed into one record by the UNION.  That should obvs never happen anyway.

source/dbo/Views/Power BI Reporting/vwUnmatchedDuplicates.sql
```diff
@@ -0,0 +1,82 @@
...
+	an2.CreationDate AS CreationDate2,
+	an2.FamilyName AS FamilyName2,
+	an2.CaseManager AS CaseManager2
+FROM AllNotifications an1
+	INNER JOIN AllNotifications an2
+		-- Notifications duplicate NHS numbers and dates of birth
+		ON an1.DateOfBirth = an2.DateOfBirth
+		AND an1.NhsNumber = an2.NhsNumber
+		-- But not the same notification twice
+		AND (an1.NtbsId IS NULL OR an2.NtbsId IS NULL OR an1.NtbsId <> an2.NtbsId)
```
aptemus, line 76: I don't think this accounts for it being the same notification but in its pre- and post-migration state. i.e. we don't want to discover that ETS 123 has the same NHS# and date of birth as NTBS 567 and it's because 567 was migrated in from 123.

aptemus (changes requested):
@benjimarshall I THINK there's a check missing to eliminate matches where it's the same notification in ETS that was then migrated into NTBS?

source/dbo/Views/Power BI Reporting/vwUnmatchedDuplicates.sql
JoshCadney: I think this is handled by the fact that migrated notifications are already filtered out of the `EtsNotifications`

JoshCadney (approved):
I agree with Nancy's suggestion around email vs username. Otherwise this looks good to me :-)

benjimarshall:
Based on the PowerBI session we want to add `COALESCE(TB service, Hospital)`, and full name instead of family name to this view, probably other things as well

benjimarshall:
Also: last treatment event (using migration db for ETS records). `COALESCE(Treatment start date, diagnosis date, notification date)` as estimated start

source/dbo/Views/Power BI Reporting/vwUnmatchedDuplicates.sql
```diff
@@ -5,40 +5,93 @@ AS
...
+		INNER JOIN [$(NTBS)].dbo.ClinicalDetails cd ON n.NotificationId = cd.NotificationId
+		INNER JOIN [$(NTBS)].dbo.HospitalDetails hd ON n.NotificationId = hd.NotificationId
+		LEFT JOIN [$(NTBS)].dbo.TreatmentEvent te ON n.NotificationId = te.NotificationId AND te.TreatmentEventType = 'TreatmentOutcome'
+		LEFT JOIN [$(NTBS)].ReferenceData.Hospital h ON hd.HospitalId = h.HospitalId
+		LEFT JOIN [$(NTBS)].ReferenceData.TbService tbs ON h.TBServiceCode = tbs.Code
+		LEFT JOIN [$(NTBS)].ReferenceData.PHEC ON tbs.PHECCode = PHEC.Code
+		LEFT JOIN [$(NTBS)].dbo.[User] u ON hd.CaseManagerId = u.Id
 		LEFT JOIN [$(migration)].dbo.EtsGroupings etsGroupings ON n.ETSID = etsGroupings.EtsId
 	WHERE n.NotificationStatus IN ('Notified', 'Closed')
+	GROUP BY n.NotificationId,
```
aptemus, line 32: Is this Group By to get the MAX to come out? I would probably do a PARTITION just to make the code easier to read (or a sub-query)

source/dbo/Views/Power BI Reporting/vwUnmatchedDuplicates.sql
```diff
@@ -10,21 +10,42 @@ WITH NtbsNotifications AS (
 		etsGroupings.GroupId AS EtsGroupId,
 		n.NotificationDate,
 		n.CreationDate,
+		COALESCE(cd.TreatmentStartDate, cd.DiagnosisDate, n.NotificationDate) AS TreatmentStartDate,
+		MAX(te.EventDate) AS LastRecordedTreatmentOutcome,
```
aptemus, line 14: I'd call this LastRecordedTreatmentOutcomeDate, as we already have a field in reporting called LastRecordedTreatmentOutcome, which holds the actual outcome (i.e. Completed) rather than the date.

aptemus (commented):
@benjimarshall couple of comments but happy to be persuaded that this is fine for now

aptemus (approved):
Looks good

PR #83, opened 2021-07-02
Ntbs 2504

aptemus:
* a new view which will create the NOIDS extract for both ETS and NTBS records
* a new table which will keep a count of the records which existed at a given point in time (i.e. each Monday when the NOIDS file is sent, count of all the records that exist at that point with a notification date on or before the Sunday before). Currently Rosanne maintains this count manually in a Word document

I do have some outstanding questions for her about what week number is used to record the counts, but this can be resolved in Power BI; as there are different ways of calculating week numbers, I have opted to record the count against a date, so the week number can be calculated later without affecting the underlying data.


source/dbo/Stored Procedures/Power BI Reporting/uspNOIDSWeeklyCount.sql
```diff
@@ -0,0 +1,23 @@
...
+CREATE PROCEDURE [dbo].[uspGenerateNOIDSWeeklyCount]
+AS
+	DECLARE @ThisWeek VARCHAR(10) =  (SELECT ISOYearWeek FROM [dbo].Calendar WHERE DateValue = CONVERT(DATE, GETUTCDATE()))
+	DECLARE @LastSunday DATE =  (SELECT MAX(DateValue) FROM [dbo].Calendar WHERE DateValue <= GETUTCDATE() AND DATENAME(dw, DateValue) = 'Sunday')
+
+	--check to see if a value has already been stored for this week
+	IF NOT EXISTS
+	(SELECT wc.DateValue 
+		FROM [dbo].[NOIDSWeeklyCount] wc
+			INNER JOIN [dbo].[Calendar] c ON c.DateValue = wc.DateValue
```
JoshCadney, line 12: This join is confusing me... it looks like `DateValue` in `Calendar` is just a date, whereas `DateValue` in `NOIDSWeeklyCount` is the exact time of a previous run, so this won't ever match anything?

adilmirzaphe (commented):
Added Lookup for LocalAuthority and updated code

benjimarshall (approved):
Looks good!

adilmirzaphe (approved):
looks good

JoshCadney (approved):
Looks good :-)

PR #84, opened 2021-07-05
Initial views for a monitoring dashboard

aptemus:
Screenshot of dashboard available on Jira ticket.
This lets us monitor activity in NTBS:
- notifications created & on demand data migrations done
- alerts created and closed
- how the Hangfire jobs did last night/last time
- user logins, user edits, record reads
- specimens unmatched

### DIFF
--- a/source/dbo/Views/Power BI Reporting/vwHangfireJobs.sql
+++ b/source/dbo/Views/Power BI Reporting/vwHangfireJobs.sql
@@ -1,0 +1,12 @@
+﻿CREATE VIEW [dbo].[vwHangfireJobs]
+	AS
+	SELECT 
+		SUBSTRING(Q1.InvocationData, 25, EndOfString-25) AS JobName, 
+		Q1.StateName, 
+		Q1.CreatedAt 
+	FROM
+		(SELECT StateName, CreatedAt, CHARINDEX(',',InvocationData, 1) AS EndOfString, InvocationData
+		FROM [$(NTBS)].[HangFire].[Job]) AS Q1
+
+  --we want the text after {"t":"ntbs_service.Jobs.UserSyncJob, ntbs-service","m":"Run","p":["Hangfire.IJobCancellationToken, Hangfire.Core"]}
+  --and before the first comma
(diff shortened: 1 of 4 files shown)


source/dbo/Views/Power BI Reporting/vwHangfireJobs.sql
```diff
@@ -0,0 +1,12 @@
+﻿CREATE VIEW [dbo].[vwHangfireJobs]
+	AS
+	SELECT 
+		SUBSTRING(Q1.InvocationData, 25, EndOfString-25) AS JobName, 
+		Q1.StateName, 
+		Q1.CreatedAt 
+	FROM
+		(SELECT StateName, CreatedAt, CHARINDEX(',',InvocationData, 1) AS EndOfString, InvocationData
+		FROM [$(NTBS)].[HangFire].[Job]) AS Q1
```
benjimarshall, line 9: I don't think this is going to play well with the on-demand migration records in Hangfire. An example of one is 

```
{"t":"ntbs_service.DataMigration.INotificationImportService, ntbs-service","m":"BulkImportByDateAsync","p":["Hangfire.Server.PerformContext, Hangfire.Core","System.Int32","System.DateTime, mscorlib","System.DateTime, mscorlib"]}
```
And an example row this produces is:
| JobName | StateName | CreatedAt |
| ---- | --------- | ------ |
| `igration.INotificationImportService` | `Deleted` | `2021-06-15 11:06:27.920` |

which I don't think was the intended result

benjimarshall (changes requested):
Just one problem I think I've spotted but otherwise it looks good!

source/dbo/Views/Power BI Reporting/vwHangfireJobs.sql
aptemus: @benjimarshall Yeah I'll add a WHERE clause to only look for records where the Invocation data includes 'ntbs_service.Jobs.'

benjimarshall (approved):
Looks good!

PR #86, opened 2021-07-09
NTBS-2500 Don't use string NHS numbers to finding duplicates

benjimarshall:
### Description
Ignore NHS numbers like 'NK' when trying to find duplicate NHS numbers across notifications.

### Testing
Tested query from view on UAT data, and invalid string NHS numbers are now excluded. This has not been deployed anywhere yet.


adilmirzaphe (approved):
Looks good

PR #87, opened 2021-07-10
Update to logic which looks for the migration job to process from NTBS

aptemus:
This did not work when we first ran it in live, because it first looks at its own table, MigrationRun, to see what is the last job number it processed. If it doesn't find one (as it would not on first execution in an environment) it looks for one with a value greater than 1. It should be zero; the first migration run has the number 1.


benjimarshall (approved):
Looks good!

JoshCadney (approved):
Looks good :-)

PR #89, opened 2021-07-15
NTBS-2532, NTBS-2558 Case manager TbService issues report

benjimarshall:
## Description

Views used by the new PowerBI report I'm making, one for the case manager issues and the other just a view on NTBS users to be used to find duplicate users.

## Testing
Tested locally, this has not been deployed anywhere yet


JoshCadney (approved):
Looks good to me :-)

PR #90, opened 2021-07-18
NTBS-2564 - new views to support report on which users appear to need…

aptemus:
… which permissions in NTBS

This is based on notifications for which they are the owner from 2018 onwards.


JoshCadney (approved):
Looks good :-)

PR #91, opened 2021-07-20
NTBS-2574: remove build warnings

aptemus:
Changed case on field names as the build warnings create noise during the release process


benjimarshall (approved):
Looks good! A couple of those look like my bad, so sorry about that!

JoshCadney (approved):
Looks good :-)

PR #92, opened 2021-07-21
Initial code for legacy lab extract

aptemus:
This view pulls together data on manual and reference lab results for both ETS and NTBS.  There are some additional columns in the final report (things like TreatmentRegion) but these items are already available in Power BI so I will add them on to the dataset there.

I've tested this against live data and it looks basically correct, but following deployment to phe-dev we will do a manual comparison with the one produced by ETS. This will not be identical, as the ETS one misses out WGS lab results, but will tell us if we've got it right.


JoshCadney (approved):
Looks good :-)

adilmirzaphe (commented):
Looks good so far. How far back is the report intended to go (currenlty limited to how far back RecordRegister goes)?

PR #93, opened 2021-07-23
NTBS-2489 Include notes from migration run outcomes without log messages

JoshCadney:
In the migration run report, when a notification does not have any log messages, get the migration notes from the outcome instead.

I've tested this against a small migration run, including one notification which fails to import because a linked notification fails validation.

### DIFF
--- a/source/dbo/Stored Procedures/Data Migration/uspGenerateMigrationResultsData.sql
+++ b/source/dbo/Stored Procedures/Data Migration/uspGenerateMigrationResultsData.sql
@@ -221,7 +221,7 @@ BEGIN TRY
 
 	UPDATE mrr
 		SET 
-			mrr.MigrationNotes = ag.[Message]
+			mrr.MigrationNotes = COALESCE(ag.[Message], ag.[Notes])
 			,mrr.MigrationResult = 
 				CASE WHEN ag.SuccessfullyMigrated = 0 THEN 'Error'
 				WHEN ag.SuccessfullyMigrated = 1 AND ag.[Message] IS NOT NULL THEN 'Data Loss'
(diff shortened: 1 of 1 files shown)


source/dbo/Stored Procedures/Data Migration/uspGenerateMigrationResultsData.sql
```diff
@@ -231,23 +231,6 @@ BEGIN TRY
...
-
-	
-	UPDATE mrr
-		SET mrr.MigrationResult = 'Error',
-		mrr.MigrationNotes = li.[Notes]
-	FROM  [dbo].[MigrationRunResults] mrr
-		INNER JOIN 
-			[$(NTBS)].[dbo].[LegacyImportNotificationOutcome] li ON li.LegacyImportMigrationRunId = mrr.MigrationRunId AND li.OldNotificationId = mrr.MigrationNotificationId
-	WHERE mrr.MigrationRunId = @MigrationRunID
-	AND mrr.MigrationResult IS NULL
```
JoshCadney, line 247: The script was previously making the assumption that for these records the `MigrationResult` has not yet been set - but this is not true - it was being set in the update above.

source/dbo/Stored Procedures/Data Migration/uspGenerateMigrationResultsData.sql
```diff
@@ -221,7 +221,7 @@ BEGIN TRY
 
 	UPDATE mrr
 		SET 
-			mrr.MigrationNotes = ag.[Message]
+			mrr.MigrationNotes = COALESCE(ag.[Message], ag.[Notes])
```
JoshCadney, line 224: I've gone for a more general approach of falling back to the Notes whenever there are no Messages. I thought that this would be most useful, in case there are other scenarios in which there are no Messages, but there are Notes.

JoshCadney: I'm assuming that in those cases we'd also want to see the Notes.

aptemus: Yes, this looks like a good approach

aptemus (approved):
Looks good 🥇

PR #94, opened 2021-07-23
NTBS-2543 Only consider submitted outcomes

JoshCadney:
Only include submitted outcomes when working out the ETS outcome in the migration run report.

Order events correctly when working out the NTBS outcome in the migration run report.


LeoCie (approved):
Looks good, was EventOrder never getting joined to before?

JoshCadney:
> Looks good, was EventOrder never getting joined to before?

Yes, exactly

PR #95, opened 2021-07-27
NTBS-2546 Stored procedures giving more detail on migration errors

JoshCadney:
Added three new stored procedures to be used by the migration run report to provide more details on potential duplicates, dubious specimens and mis-matched outcomes


source/dbo/Stored Procedures/Data Migration/uspMigrationSpecimensMatchesToReview.sql
```diff
@@ -1,13 +1,16 @@
 ﻿CREATE PROCEDURE [dbo].[uspMirgationSpecimensMatchesToReview]
```
JoshCadney, line 1: For this procedure, I have replaced an existing stored procedure, because I think we won't need the old one once we've made the changes to the report, but I'm not totally certain, so it's possible we'll need to restore this.

source/dbo/Stored Procedures/Data Migration/uspMigrationOutcomeMismatch.sql
```diff
@@ -0,0 +1,38 @@
+﻿/*
+	Note that we treat some outcomes as being equal even when they are actually different in NTBS and ETS.
+	For more details see uspMigrationMisMatchedOutcomes.sql.
+*/
+
+CREATE PROCEDURE [dbo].[uspMigrationOutcomeMismatch]
```
JoshCadney, line 6: There is a similar sproc that already exists (`uspMigrationMisMatchedOutcomes.sql`) but that one's being used to aggregate the number of mismatched outcomes, whereas this one is showing more detail. I don't know if we'll still want the old one once we've updated the report?

source/dbo/Stored Procedures/Data Migration/uspMigrationOutcomeMismatch.sql
```diff
@@ -0,0 +1,38 @@
...
+		UNION
+		SELECT 'TreatmentStart' AS EventName, 2 AS 'OrderBy'
+		UNION
+		SELECT 'TransferOut' AS EventName, 3 AS 'OrderBy'
+		UNION
+		SELECT 'TransferIn' AS EventName, 4 AS 'OrderBy'
+		UNION
+		SELECT 'TransferRestart' AS Eventname, 5 AS 'OrderBy'
+		UNION
+		SELECT 'TreatmentOutcome' AS EventName, 6 AS 'OrderBy')
```
JoshCadney, line 20: This trick is duplicated in `uspGenerateMigrationResultsData.sql` - we could commonise the two, by joining to a view of notifications and their most recent treatment event. Do you think this would be worthwhile?

source/dbo/Stored Procedures/Data Migration/uspMigrationOutcomeMismatch.sql
```diff
@@ -0,0 +1,38 @@
...
+	SELECT DISTINCT
+		mrr.MigrationNotificationId		AS 'MigrationNotificationId',
+		mrr.LegacyETSId					AS 'EtsId',
+		mrr.NTBSNotificationId			AS 'NtbsId',
+		mrr.EtsTreatmentOutcome			AS 'EtsOutcome',
+		mrr.NTBSTreatmentOutcome		AS 'NtbsOutcome',
+		FIRST_VALUE(te.EventDate) OVER (PARTITION BY NotificationId ORDER BY EventDate DESC, OrderBy DESC)			AS 'FinalEventDate',
+		FIRST_VALUE(te.TreatmentEventType) OVER (PARTITION BY NotificationId ORDER BY EventDate DESC, OrderBy DESC)	AS 'FinalEventType'
+	FROM [dbo].[MigrationRunResults] mrr
+		INNER JOIN [$(NTBS)].[dbo].[TreatmentEvent] te ON te.NotificationId = mrr.NTBSNotificationId
```
JoshCadney, line 31: I've used an inner join here, because I think every notification is guaranteed to have at least one treatment event, but looking at this now, perhaps a left join would have been safer

source/dbo/Stored Procedures/Data Migration/uspMigrationOutcomeMismatch.sql
aptemus: We can do the aggregation in SSRS from the detailed data so we shouldn't need both.

source/dbo/Stored Procedures/Data Migration/uspMigrationDuplicateAlerts.sql
```diff
@@ -0,0 +1,18 @@
+﻿CREATE PROCEDURE [dbo].[uspMigrationDuplicateAlerts]
+	@MigrationRun INT = NULL
+AS
+	SELECT
+		mrr.MigrationNotificationId												AS 'MigrationNotificationId',
+		mrr.LegacyETSId															AS 'EtsId',
+		mrr.NTBSNotificationId													AS 'NtbsId',
+		mrr.NotificationDate													AS 'NotificationDate',
+		tbService.[Name]														AS 'TbService',
```
aptemus, line 9: The TB service name is already stored in Migration Run Results, so there's no need to calculate it again here.

source/dbo/Stored Procedures/Data Migration/uspMigrationOutcomeMismatch.sql
aptemus: The NTBS and ETS treatment outcomes are stored in Migration Run Results, there's no need to recalculate them here.  What I was looking for was a list of the treatment events from NTBS with their dates, as this will normally illustrate why the outcomes are mismatched.

source/dbo/Stored Procedures/Data Migration/uspMigrationOutcomeMismatch.sql
aptemus: Yeah I would go for a left join just to be safe. An inner join _should_ never fail here but you never know.

source/dbo/Stored Procedures/Data Migration/uspMigrationSpecimensMatchesToReview.sql
aptemus: I notice that the SP has a typo in the name!  We might as well correct that at the same time.

source/dbo/Stored Procedures/Data Migration/uspMigrationSpecimensMatchesToReview.sql
```diff
@@ -1,13 +1,16 @@
...
+		mrr.MigrationNotificationId							AS 'MigrationNotificationId',
+		mrr.LegacyETSId										AS 'EtsId',
+		mrr.NTBSNotificationId								AS 'NtbsId',
+		mdsm.ReferenceLaboratoryNumber						AS 'ReferenceLaboratoryNumber',
+		esm.SpecimenDate									AS 'SpecimenDate',
+		mrr.NotificationDate								AS 'NotificationDate',
+		dbo.ufnGetTreatmentEndDate(mrr.NTBSNotificationId)	AS 'TreatmentEndDate',
+		mdsm.MigrationNotes									AS 'MigrationNotes'
+	FROM [dbo].[MigrationRunResults] mrr
+		INNER JOIN [dbo].[MigrationDubiousSpecimenMatches] mdsm ON mdsm.EtsId = mrr.LegacyETSId
```
aptemus, line 14: Shouldn't this be vwMigrationDubiousSpecimenMatches?

aptemus (changes requested):
@JoshCadney might be worth having a quick catch-up esp around the outcome stuff.

source/dbo/Stored Procedures/Data Migration/uspMigrationOutcomeMismatch.sql
JoshCadney: I thought the idea here was to include the most recent treatment event (i.e. the one from which the outcome is taken)... I didn't realise we wanted to include all the treatment events.

source/dbo/Stored Procedures/Data Migration/uspMigrationDuplicateAlerts.sql
JoshCadney: Ah, so it is! Thanks :-)

source/dbo/Stored Procedures/Data Migration/uspMigrationSpecimensMatchesToReview.sql
JoshCadney: `vwMigrationDubiousSpecimenMatches` aggregates `MigrationDubiousSpecimenMatches` by notification, so we need to go to the latter since we want one row per specimen rather than one row per notification.

source/dbo/Stored Procedures/Data Migration/uspGenerateMigrationResultsData.sql
```diff
@@ -124,13 +124,10 @@ BEGIN TRY
...
-	LEFT OUTER JOIN  [$(ETS)].[dbo].[TreatmentOutcomeTwentyFourMonth] tr24 ON tr24.Id = n.TreatmentOutcomeTwentyFourMonthId
-	LEFT OUTER JOIN  [$(ETS)].[dbo].[TreatmentOutcome36Month] tr36 ON tr36.Id = n.TreatmentOutcome36MonthId
-	WHERE mrr.MigrationRunId = @MigrationRunID
-	AND tr12.Submitted = 1
-	AND tr24.Submitted = 1
-	AND tr36.Submitted = 1;
+	LEFT OUTER JOIN  [$(ETS)].[dbo].[TreatmentOutcome] tr12 ON tr12.Id = n.TreatmentOutcomeId AND tr12.Submitted = 1
+	LEFT OUTER JOIN  [$(ETS)].[dbo].[TreatmentOutcomeTwentyFourMonth] tr24 ON tr24.Id = n.TreatmentOutcomeTwentyFourMonthId AND tr24.Submitted = 1
+	LEFT OUTER JOIN  [$(ETS)].[dbo].[TreatmentOutcome36Month] tr36 ON tr36.Id = n.TreatmentOutcome36MonthId AND tr36.Submitted = 1
+	WHERE mrr.MigrationRunId = @MigrationRunID;
```
JoshCadney, line 130: This is a fix for a bug in NTBS-2543

JoshCadney (commented):
This is back up for review, with most of the review comments resolved. The one thing I haven't looked into yet is getting rid of the aggregated mismatched outcomes view and doing the aggregation in SSRS.

I'm not expecting anyone to review the `.rdl` file (that will be assessed in QA).

aptemus (approved):
Looks good 👍

source/dbo/Stored Procedures/Data Migration/uspMigrationOutcomeMismatch.sql
JoshCadney: This is now done

JoshCadney:
I've now added the final change - removing the now redundant sproc for mis-matched outcomes and using a single dataset for both the aggregated and detailed tables.

LeoCie (approved):
Looks good to me!

PR #96, opened 2021-07-27
NTBS-2588 incorrect date appears in report footer for NTBS

aptemus:
The code had not been updated after the audit database was restructured, so it was looking in the 'EntityType' column for the value 'Notification' when it should be looking in the 'RootEntity' column.  This is having a minor effect in live, depending on what kind of updates are performed the day before.


JoshCadney (approved):
Looks good :-)

PR #97, opened 2021-07-29
NTBS-2570: updates to NOIDS view

aptemus:
We needed to correct two problems:
* cases were getting dropped from the view if they had no sites of disease at all in ETS (not possible in NTBS)
* a new source for the particular kind of local authority codes used by NOIDS, which meant moving away from our previous look-up table and adding a new column to the reduced_postcode_file table in the geography database.  (This has now been populated in all environments).


JoshCadney (approved):
Looks good :-)

PR #101, opened 2021-08-03
NTBS-2335: Fix call to started treatment in NTBS

LeoCie:
Reflect the change made in the NTBS database in the reporting code. The DACPAC will also need to be updated but from previous experience we know that needs to be done once the DB has been updated on our azure environment and not locally.


JoshCadney (approved):
Looks good :-)

PR #102, opened 2021-08-12
NTBS-1141 Missing mappings report

benjimarshall:
## Description
* Refactor and add the ETS user ID to the `vwEtsUserPermissionMembership` view, so we can join on that directly for our code here
* Add a view to find missing hospital mappings
* Add a view to find antibiotics that were not mapped during import to `LabBase2` (so `AntibioticCode = 'New'`)
* Add a view to find antibiotic codes that were mapped during import to `LabBase2`, but not known in the `specimen-matching` mapping.

## Testing
Tested that it builds and deploys locally, and have tested the queries in the views against live data and they behave as expected


source/dbo/Views/Power BI Reporting/vwMissingAntibioticMappings.sql
```diff
@@ -0,0 +1,15 @@
...
+
+SELECT su.AntibioticCode,
+	COUNT(DISTINCT a.ReferenceLaboratoryNumber) AS NumberOfSamples,
+	MIN(a.AuditCreate) AS DateFirstIntroduced,
+	MAX(a.AuditCreate) AS LastSeen
+FROM [$(Labbase2)].dbo.Susceptibility su
+	INNER JOIN [$(Labbase2)].dbo.Anonymised a ON a.LabDataID = su.LabDataID
+WHERE su.AntibioticCode NOT IN (SELECT AntibioticCode FROM [$(NTBS_Specimen_Matching)].dbo.AntibioticMapping)
+	-- Don't warn about not being able to map things that aren't antibiotics
+	AND su.AntibioticCode NOT IN ('New', 'Gynaecological', 'Other tissues')
```
benjimarshall, line 12: As agreed on the ticket we are only ignoring the `AntibioticCode`s that aren't actually antibiotics. The ones that we haven't mapped because we don't care about them and they haven't come up in a TB case yet are allowed to come through, in case they ever are seen on a case with `IsAtypicalOrganismRecord = 0`

source/dbo/Views/Power BI Reporting/vwMissingHospitalMappings.sql
```diff
@@ -0,0 +1,61 @@
...
+AS
+
+-- LTBR hospital is not in migration mapping
+SELECT CONCAT(dp_PatientID, '-', dp_DiseasePeriod) AS OldNotificationId,
+	'LTBR' AS [Source],
+	dp_NotifiedDate AS NotificationDate,
+	CAST(dp_NotifyingClinicID AS NVARCHAR(50)) AS HospitalId,
+	lh_HospitalName COLLATE Latin1_General_CI_AS AS HospitalName,
+	residencePhec.[Name] AS ResidenceRegion,
+	CAST(NULL AS NVARCHAR) AS TreatmentRegion
```
benjimarshall, line 11: I couldn't think of a good way of trying to extract this from LTBR, if anyone can think of a way do let me know otherwise they can just be left as `NULL`

source/dbo/Views/Power BI Reporting/vwNewAntibioticCodes.sql
```diff
@@ -0,0 +1,15 @@
...
+	a.ReferenceLaboratory,
+	COUNT(DISTINCT a.ReferenceLaboratoryNumber) AS NumberOfSamples,
+	MIN(a.AuditCreate) AS FirstIntroduced,
+	MAX(a.AuditCreate) AS LastSeen
+FROM [$(Labbase2)].dbo.Susceptibility su
+	INNER JOIN [$(Labbase2)].dbo.Anonymised a ON a.LabDataID = su.LabDataID
+WHERE su.AntibioticCode = 'New'
+	AND a.IsAtypicalOrganismRecord = 0
+	AND a.MergedRecord = 0
+GROUP BY su.LabAntibiotic, a.ReferenceLaboratory
```
benjimarshall, line 15: This wasn't specified in the ticket but it felt right to break out the missing antibiotic issue into two views: one where we are missing the mapping (the other view to this one), and one where it wasn't mapped properly during `LabBase2` import. These are a rough guess of what fields that will be useful when debugging this but let me know if you have any suggestions

benjimarshall (commented):
Just a few things to flag up

benjimarshall:
And one other thing I spotted when doing this ticket: there are two LTBR hospitals that have a corresponding `NULL` NTBS hospital ID. [See here](https://github.com/publichealthengland/ntbs-data-migration/blob/13c3b465909a87ea1756b3bc68abf8d0ea24abfe/ntbs-data-migration/dbo/Stored%20Procedures/uspSeedStaticMappingTables.sql#L1302). I can't find any instances of these hospitals being used in cases we might import (it's not in `migration.dbo.LTBR_DiseasePeriod`), but if a case with one of these hospitals does appear we won't be able to import it (it will fail validation). Currently this view will warn us about this issue, but it might be worth fixing in migration.

For context, in ETS migration we coalesce the possible hospital IDs and fall back to an unknown hospital GUID (that has an ID, TB Service, etc.). You can see this [here](https://github.com/publichealthengland/ntbs-data-migration/blob/13c3b465909a87ea1756b3bc68abf8d0ea24abfe/ntbs-data-migration/dbo/Views/20_migration_views/EtsNotifications.sql#L13). Maybe we want to put the same in for LTBR in migration?

source/dbo/Views/Power BI Reporting/vwMissingHospitalMappings.sql
```diff
@@ -0,0 +1,61 @@
...
+SELECT CONCAT(dp_PatientID, '-', dp_DiseasePeriod) AS OldNotificationId,
+	'LTBR' AS [Source],
+	dp_NotifiedDate AS NotificationDate,
+	CAST(dp_NotifyingClinicID AS NVARCHAR(50)) AS HospitalId,
+	lhm.LtbrName AS HospitalName,
+	residencePhec.[Name] AS ResidenceRegion,
+	CAST(NULL AS NVARCHAR) AS TreatmentRegion
+FROM [$(migration)].dbo.LTBR_DiseasePeriod dp
+	INNER JOIN [$(migration)].dbo.LtbrAddresses a ON a.OldNotificationId = CONCAT(dp_PatientID, '-', dp_DiseasePeriod)
+	LEFT JOIN [$(migration)].dbo.LtbrHospitalsMapping lhm ON dp_NotifyingClinicID = lhm.LtbrId
```
JoshCadney, line 33: Could we make this an inner join, since the cases where the `dp_NotifyingClinicID` is not found in this table are already included in the select above?

JoshCadney (commented):
One possible suggestion, but otherwise looks good to me

benjimarshall:
@aptemus would you be able to review this when you get the chance?

benjimarshall:
@adilmirzaphe Nancy has asked if you could do this review instead, because you're more familiar with the data sources, if that's ok?

adilmirzaphe (approved):
Looks good to me. :-)

PR #103, opened 2021-08-26
NTBS-2659: Improve migration results report

LeoCie:
Adding some extra data to the procedures which are used in the migration run reports. 
Added ETS ID and 'Demographics match' of the duplicate notification in the case of a duplicate alert - and made it so that it only outputs one row per duplicate.
Made the dubious specimen match return all the notifications which were part of the dubious match, not only the notification which was part of the migration. Also returned the status of the notifications in ETS. 

The only changes to the actual report are: Adding in the extra fields and now grouping the specimens by reference lab number instead of by notification

### DIFF
--- a/source/dbo/Stored Procedures/Data Migration/uspMigrationDuplicateAlerts.sql
+++ b/source/dbo/Stored Procedures/Data Migration/uspMigrationDuplicateAlerts.sql
@@ -1,21 +1,44 @@
 ﻿CREATE PROCEDURE [dbo].[uspMigrationDuplicateAlerts]
 	@MigrationRun INT = NULL
 AS
+	WITH DuplicatePairs AS
+	(
+		SELECT DISTINCT
+		CASE WHEN mrr.NotificationDate < duplicate.NotificationDate THEN mrr.NTBSNotificationId ELSE duplicate.NotificationId END AS 'OriginalNotification',
+		CASE WHEN mrr.NotificationDate > duplicate.NotificationDate THEN mrr.NTBSNotificationId ELSE duplicate.NotificationId END AS 'DuplicateNotification'
+		FROM [dbo].[MigrationRunResults] mrr
+			INNER JOIN [$(NTBS)].[dbo].[Alert] a on a.NotificationId = mrr.NTBSNotificationId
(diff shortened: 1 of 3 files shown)


source/dbo/Stored Procedures/Data Migration/uspMigrationDuplicateAlerts.sql
```diff
@@ -4,41 +4,51 @@ AS
...
-		LEFT JOIN [$(NTBS)].[dbo].[Patients] originalp on originalp.NotificationId = original.NotificationId
-		LEFT JOIN [$(NTBS)].[dbo].[HospitalDetails] originalHd on originalHd.NotificationId = original.NotificationId
-		LEFT JOIN [$(NTBS)].[ReferenceData].[TbService] originalTbService on originalTbService.Code = originalHd.TBServiceCode
-		LEFT JOIN [$(NTBS)].[dbo].[Notification] duplicate on duplicate.NotificationId = pairs.DuplicateNotification
-		LEFT JOIN [$(NTBS)].[dbo].[Patients] duplicatep on duplicatep.NotificationId = duplicate.NotificationId
-		LEFT JOIN [$(NTBS)].[dbo].[HospitalDetails] duplicateHd on duplicateHd.NotificationId = duplicate.NotificationId
-		LEFT JOIN [$(NTBS)].[ReferenceData].[TbService] duplicateTbService on duplicateTbService.Code = duplicateHd.TBServiceCode
\ No newline at end of file
+		LEFT JOIN [dbo].[MigrationRunResults] mrr ON mrr.NTBSNotificationId = pairs.FirstNotification
+		LEFT JOIN [$(NTBS)].[dbo].[Notification] firstNot on firstNot.NotificationId = pairs.FirstNotification
```
JoshCadney, line 47: Minor thing, but I don't really like `Not` as an abbreviation of `Notification` as it has a meaning of its own. I probably would have gone with `firstNotification` and then just `firstPatient` etc. That said, SQL table names are often less meaningful than this!

JoshCadney (approved):
A minor markup suggested, but I'm happy for you to merge.

PR #104, opened 2021-09-02
NTBS-2659: Fix migration report stored procs

LeoCie:
I hadn't thought about the fact that MigrationRunResults will have all the old runs in, so was joining all cases of a notification. Now join on the MigrationRunId also.


JoshCadney (approved):
Looks good :-)

PR #105, opened 2021-09-07
NTBS-2659: Update migration results report

LeoCie:
Updated the rdl file for the migration results report. 
The update consisted of removing some unwanted fields and switching the order of notifications in the duplicate table.


JoshCadney (approved):
Looks good :-)

PR #106, opened 2021-09-07
NTBS-2589: Add migration version to migration run table

LeoCie:
Add the migration version to the migration run table, I'll add the rdl file to this once I edit the report.

### DIFF
--- a/source/dbo/Stored Procedures/Data Migration/uspConfirmDatabaseVersions.sql
+++ b/source/dbo/Stored Procedures/Data Migration/uspConfirmDatabaseVersions.sql
@@ -3,7 +3,7 @@
 	
 AS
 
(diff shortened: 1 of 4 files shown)


JoshCadney (approved):
Looks good :-)

source/dbo/Stored Procedures/Data Migration/uspConfirmDatabaseVersions.sql
```diff
@@ -3,7 +3,7 @@
 	
 AS
 
-	SELECT EtsDate, LabbaseDate, LtbrDate, AppVersion FROM
+	SELECT EtsDate, LabbaseDate, LtbrDate, AppVersion, MigrationVersion FROM
```
LeoCie, line 6: re-requested review as I changed this stored proc, and the report.

PR #107, opened 2021-09-07
NTBS-2677: Edit report footer text

LeoCie:
Remove the word "database" from the reporting footer. I have also updated the identifiable dataset pbix as part of this ticket to no longer add the Power BI reporting version to the footer. That change is on the NTBS reports on azure


JoshCadney (approved):
Looks good :-)

PR #108, opened 2021-09-08
NTBS-2676: Add columns to reporting views needed in report

LeoCie:
The new TB service report needed a couple of extra columns from these views.


JoshCadney (approved):
Looks good to me :-)

PR #109, opened 2021-09-13
NTBS-2700: summarise results of sputum testing

aptemus:
Sputum testing is normally done for all pulmonary cases and in some regions it is a KPI which is reviewed at cohort review. We have been asked to add a field to the NTBS [enhanced] line list that summarises the results of all sputum tests for the case.

There is an outstanding question about whether this should exclude culture results (probably) or line probe assay (probably not).  Hence the TODO comments and commented out line in the code.  If we don't hear from the Yorkshire team tomorrow, I will leave it as written and we can hot fix a change if required.

I can't really develop the Power BI report change until I can merge the SQL in.


JoshCadney (approved):
Looks good to me :-)

PR #110, opened 2021-09-15
NTBS-2700 clarify test types required

aptemus:
Should be only smear and PCR, which in ETS equates to Microscopy and Molecular Amplification


JoshCadney (approved):
Looks good :-)

PR #111, opened 2021-09-16
NTBS-2705: Add requested case manager to multi system transfers

LeoCie:
Power BI reports edited, uploaded to one drive and deployed to azure test.
I have also added the south east transition date into the table which will pick up the transfers.


JoshCadney (approved):
Looks good :-)

It looks like adding the region's cut over date is something that has slipped through the net of our process. Could you add a note to the "Perform data migration" page in confluence stating that this should have happened before the live migration?

PR #113, opened 2021-09-28
NTBS-2647: Set started treatment to yes if treatment start date populated

LeoCie:
Changed ETS case record to set started treatment to yes if treatment start date populated.
Not needed for the NTBS case records, as we only allow the input of a date if they have set Started treatment to Yes.


JoshCadney (approved):
Looks good :-)

PR #114, opened 2021-09-29
NTBS-2742 fix for ets 36 month end of treatment date

adilmirzaphe:
@aptemus , please review/qa this


aptemus (changes requested):
@adilmirzaphe I did tell @JoshCadney we'd delete the offending line of code - commenting it out looks like we think we might need it again in the future, which we definitely don't.

JoshCadney:
We can now delete this branch (and corresponding PR) as the code is on the live branch

PR #116, opened 2021-09-30
NTBS-2731: Order disease sites in case records to match NTBS

LeoCie:
Added ordering to the sites of disease in Record_CaseData so it mirrors the ordering on the clinical details page


JoshCadney (approved):
Looks good :-)

PR #118, opened 2021-10-05
NTBS-2405 Changes from final testing

JoshCadney:
- Update the publish profiles to include the domain so that they can be accessed from the OpenShift cluster
- Include the master dacpac in the source code so that it can be used during deployment


LeoCie (approved):
Looks good!

PR #121, opened 2021-10-18
NTBS-2785: SQL for Cohort Review

LeoCie:
I've run all new queries on the devbox, and most don't take longer than a couple of seconds. The only query that we may need to worry about is the manual test results query for ETS case data, this takes ~20-40s but it is only run once so it may not be a huge issue.

### DIFF
--- a/source/dbo/Stored Procedures/Power BI Reporting/uspGenerateEtsCaseRecord.sql
+++ b/source/dbo/Stored Procedures/Power BI Reporting/uspGenerateEtsCaseRecord.sql
@@ -7,6 +7,30 @@ BEGIN TRY
 		[Description] nvarchar(2000)
 	);
 
+	DECLARE @TempManualTestResult TABLE
+	(
+		NotificationId int,
+		ManualTestTypeId int,
+		[Result] nvarchar(50)
+	);
+
+	DECLARE @TempSocialContextVenues TABLE
+	(
+		NotificationId int,
+		[VenueCount] int,
+		[Description] nvarchar(2000)
+	);
+
+	DECLARE @TempMBovDetails TABLE
+	(
+		NotificationId int,
+		[IsMBovis] bit,
+		[HasAnimalExposure] varchar(10),
+		[HasUnpMilkConsumption] varchar(10),
+		[HasExposureToKnownCase] varchar(10),
+		[HasOccupationExposure] varchar(10)
+	);
+
 	INSERT INTO @TempDiseaseSites
 	SELECT n.TuberculosisEpisodeId, STRING_AGG(sites.[Name], N', ') WITHIN GROUP (ORDER BY ord.OrderIndex) AS [Description]
 		FROM RecordRegister rr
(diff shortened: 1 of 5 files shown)


LeoCie:
@aptemus that does help, but it doesn't give the time period. Thought it might be useful to say "Yes - Current" or "Yes - in last 5 years". It isn't a lot of logic to add a string to the end of another, but wondered if we'd want to do it here

source/dbo/Stored Procedures/Power BI Reporting/uspGenerateNtbsCaseRecord.sql
```diff
@@ -15,6 +29,25 @@ BEGIN TRY
 		WHERE rr.SourceSystem = 'NTBS'
 		GROUP BY rr.NotificationId;
 
+	INSERT INTO @TempManualTestResult
+	SELECT DISTINCT rr.NotificationId, ManualTestTypeId,
```
aptemus, line 33: You could use a where clause to exclude Chest x-ray (and the upcoming Chest CT) test types here, as you won't be trying to summarise them for this anyway?

source/dbo/Stored Procedures/Power BI Reporting/uspGenerateReportingRecords.sql
```diff
@@ -100,6 +101,7 @@ BEGIN TRY
 		,p.NhsNumber										AS NhsNumberToLookup --this will later be reformatted if valid
 		,p.GivenName										AS Forename
 		,p.FamilyName										AS Surname
+		,LEFT(p.GivenName,1)+'.'+LEFT(p.FamilyName,1)+'.'   AS Initials
```
aptemus, line 104: Not sure I would have gone as far as storing the initials in SQL server but as you have, I would just store the initials with no dots between them.

source/dbo/Stored Procedures/Power BI Reporting/uspGenerateReportingRecords.sql
```diff
@@ -125,6 +128,7 @@ BEGIN TRY
 		,p.NhsNumber										AS NhsNumberToLookup --this will later be reformatted if valid
 		,p.Forename											AS GivenName
 		,p.Surname											AS FamilyName
+		,LEFT(p.Forename,1)+'.'+LEFT(p.Surname,1)+'.'		AS Initials
```
aptemus, line 131: Bit hard to follow why this is in twice but this type of transform that can be done on all records can just be done once at the end of the case data SP, in uspGenerateReportingCaseData

source/dbo/Tables/Power BI Reporting/Record_CaseData.sql
```diff
@@ -52,12 +52,15 @@
 	[MdrTreatmentDate] date NULL,
 	[EnhancedCaseManagement] [varchar] (10) NULL,
 	[EnhancedCaseManagementLevel] [nvarchar] (10) NULL,
+	[FirstPresentationSetting] [nvarchar] (25) NULL,
```
aptemus, line 55: I don't think this will be long enough - the free text healthcare setting description field is 100 characters in the database.

source/dbo/Stored Procedures/Power BI Reporting/uspGenerateNtbsCaseRecord.sql
LeoCie: Yep added where clause

source/dbo/Stored Procedures/Power BI Reporting/uspGenerateEtsCaseRecord.sql
```diff
@@ -18,6 +42,55 @@ BEGIN TRY
...
+				WHEN 3 THEN 'Awaiting'
+				ELSE 'No result'
+			END) OVER (PARTITION BY rr.NotificationId, mttm.NtbsId ORDER BY
+			CASE lr.Result
+				WHEN 0 THEN 2
+				WHEN 1 THEN 1
+				WHEN 2 THEN 3
+				WHEN 3 THEN 3
+				ELSE 4
+			END) AS [Result]
```
LeoCie, line 60: @benjimarshall A nicer way of doing this? Or is this the best we're gonna get?

PR #124, opened 2021-10-29
Ntbs 2823 ETS on linked server

adilmirzaphe:
Added changes to connect ETS to linked server. Only UAT publish profile has been updated as just scoping for now.


adilmirzaphe:
Ah, added now

aptemus (changes requested):
@adilmirzaphe I don't think creating a new SQLCmdVariable is needed, is it? The ETS variable just needs to contain SQLClosColStudy.ets as its value? If we *do* need it, it can't be called 'OtherServer' as it will be too confusing for another team trying to maintain the code.

PR #127, opened 2021-11-03
NTBS-2788: Add no result coalesce for chest CT

LeoCie:
Noticed while I was doing the power BI changes.


JoshCadney (approved):
Looks good :-)

PR #128, opened 2021-11-11
NTBS-2788: Update NTBS dacpac and add hpt field

LeoCie:
Updated the DACPAC and changed where CaseData was getting the MDR fields from. 
Added the HPT reference number field


JoshCadney (approved):
Looks good :-)

PR #129, opened 2021-11-14
Ntbs 2850

aptemus:
Add support to the migration results report to source the outcome from LTBR for LTBR primary records, and find mismatches that **aren't** where the outcome has been deliberately mapped from one LTBR value (e.g. completed with culture conversion) to one NTBS value (e.g. Cured).


source/dbo/Tables/Data Migration/MigrationRunResults.sql
```diff
@@ -11,13 +11,15 @@
 	[SourceSystem] [nvarchar](10)  NULL,
 	[LegacyHospitalId] [varchar](50)  NULL,
 	[LegacyHospitalName] [nvarchar](255)  NULL,
-	[EtsTreatmentOutcome] [varchar] (30) NULL,
+	[LegacyOutcomeId] [int] NULL,
+	[EtsTreatmentOutcome] [varchar] (50) NULL,
```
JoshCadney, line 15: It may not be worth it, but do you think we should rename this field `LegacyTreatmentOutcome`?

JoshCadney (approved):
One question, but looks good :-)

source/dbo/Tables/Data Migration/MigrationRunResults.sql
aptemus: We should, but it's too much hassle at this stage given it then affected the report as well.  I decided not to make things worse by introducing a field called 'EtsOutcomeId' when it patently wasn't, but this report is nearly at the end of its natural life now anyway.

PR #132, opened 2021-11-30
NTBS-2813: Add no result test result

LeoCie:
Also I found a bug, where we hadn't added excluded chest CTs from the legacy lab extract.
Nothing in GenerateCaseData needs to be changed, as we coalesce on "No result", and our query will not pick up "NoResult" (the new result) results. So these will end up as "No result" which is the desired outcome.


aptemus (approved):
@LeoCie Looks good

PR #135, opened 2021-12-09
NTBS-2889: Add drug resistance profiles view

LeoCie:


aptemus (approved):
@LeoCie looks good - I would have done 'SELECT 'No result' UNION SELECT 'INH resistant' etc but your way works fine!

PR #136, opened 2021-12-13
NTBS-2765: Remove unneeded columns from missing items view

LeoCie:
The power BI report just joins to the table it already has!


JoshCadney (approved):
Looks good to me :-)

PR #137, opened 2021-12-13
NTBS-2935: Remove now unused view

LeoCie:


JoshCadney (approved):
Looks good :-) Are we bringing this into 1.11?

JoshCadney:
> Looks good :-) Are we bringing this into 1.11?

Update: I spoke to Nancy, and we think holding off until 1.12 is better

PR #138, opened 2021-12-20
Include additional year in reporting

aptemus:
This will ensure 2018 is not dropped at the start of 2022.


JoshCadney (approved):
Looks good :-)

PR #140, opened 2022-01-05
NTBS-2944: Fix service responsible fields in case data

LeoCie:
Only generate a "service responsible at X months" if there is an outcome at X months (for 24 and 36, 12 is always calculated).
Brought the logic into it's own function.
Fixed the case where there is a transfer on the day of the outcome.

### DIFF
--- a/source/dbo/Functions/Reusable Notification/ufnGetServiceResponsible.sql
+++ b/source/dbo/Functions/Reusable Notification/ufnGetServiceResponsible.sql
@@ -1,6 +1,6 @@
 ﻿CREATE FUNCTION [dbo].[ufnGetServiceResponsible]
 (
-	@TimePeriod int,
+	@NumberOfYears int,
 	@NotificationId int,
 	@StartDate DATETIME,
 	@DefaultService NVARCHAR(150)
@@ -13,19 +13,10 @@ BEGIN
 	SET @ReturnValue = 
 		COALESCE (
 			(
-			SELECT TOP 1 trans.TbServiceName
-			FROM AllTransfers trans
-			WHERE trans.NotificationId = @NotificationId
-				AND trans.EventDate = DATEADD(DAY, -1, DATEADD(YEAR, @TimePeriod, @StartDate))
-				AND trans.TransferType = 'TransferIn'
-			ORDER BY EventDate DESC
-			),
-			(
-			SELECT TOP 1 trans.TbServiceName
-			FROM AllTransfers trans
-			WHERE trans.NotificationId = @NotificationId
-				AND trans.EventDate >= DATEADD(DAY, -1, DATEADD(YEAR, @TimePeriod, @StartDate))
-				AND trans.TransferType = 'TransferOut'
+			SELECT TOP 1 tout.TbServiceName
+			FROM TransfersOut tout
+			WHERE tout.NotificationId = @NotificationId
+				AND tout.EventDate > DATEADD(DAY, -1, DATEADD(YEAR, @NumberOfYears, @StartDate))
 			ORDER BY EventDate
 			),
 			@DefaultService
--- a/source/dbo/Functions/Reusable Notification/ufnGetServiceResponsible.sql
+++ b/source/dbo/Functions/Reusable Notification/ufnGetServiceResponsible.sql
@@ -1,0 +1,37 @@
+﻿CREATE FUNCTION [dbo].[ufnGetServiceResponsible]
+(
+	@TimePeriod int,
+	@NotificationId int,
+	@StartDate DATETIME,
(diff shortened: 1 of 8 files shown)


aptemus (commented):
@LeoCie I've added some comments but let's see with @JoshCadney thinks.

source/dbo/Functions/Reusable Notification/ufnGetServiceResponsible.sql
```diff
@@ -0,0 +1,37 @@
+﻿CREATE FUNCTION [dbo].[ufnGetServiceResponsible]
+(
+	@TimePeriod int,
```
JoshCadney, line 3: I think `NumberOfYears` would be better than `TimePeriod` as a name for this variable

source/dbo/Functions/Reusable Notification/ufnGetServiceResponsible.sql
```diff
@@ -0,0 +1,37 @@
...
+			ORDER BY EventDate DESC
+			),
+			(
+			SELECT TOP 1 trans.TbServiceName
+			FROM AllTransfers trans
+			WHERE trans.NotificationId = @NotificationId
+				AND trans.EventDate >= DATEADD(DAY, -1, DATEADD(YEAR, @TimePeriod, @StartDate))
+				AND trans.TransferType = 'TransferOut'
+			ORDER BY EventDate
+			),
```
JoshCadney, line 30: I don't think we need these two cases, I think it's equivalent to:
```
SELECT TOP 1 trans.TbServiceName
			FROM AllTransfers trans
			WHERE trans.NotificationId = @NotificationId
				AND trans.EventDate > DATEADD(DAY, -1, DATEADD(YEAR, @TimePeriod, @StartDate))
				AND trans.TransferType = 'TransferOut'
			ORDER BY EventDate
```
i.e. if there's a transfer on the day then the transfer in should match the transfer out of the next transfer (and if there isn't another transfer then it should match the current TB service)

JoshCadney: Having thought about it a little more, they won't be equivalent if the transfers on the record aren't consistent with the current TB service, but in that case I think it's better to prefer the current TB service anyway.

source/dbo/Stored Procedures/Power BI Reporting/uspGenerateRecordOutcome.sql
```diff
@@ -40,48 +40,38 @@ BEGIN TRY
...
 		TreatmentOutcome12months = po1.OutcomeValue, 
 		TreatmentOutcome24months = po2.OutcomeValue, 
 		TreatmentOutcome36months = po3.OutcomeValue,
 		NotifyingTbService =
-		CASE
-			WHEN cd.NotificationId IN (SELECT NotificationId FROM transfersOut)
-				THEN (SELECT TOP 1 tout.[Name] FROM transfersOut tout ORDER BY EventDate)
-			ELSE cd.TbService
-		END,
+		COALESCE(
```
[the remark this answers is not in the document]
JoshCadney: Yes, I suppose it's true that "notifying TB service" isn't always equivalent to "TB service responsible at 0 years" if the case is transferred after notification but before the treatment start date.

I agree that a subquery would be slightly better.

JoshCadney (changes requested):

JoshCadney (commented):
I think this looks good, but I also think we'll want to make changes following NTBS-2932 so that the notifying clinic/case manager are taken from the initial event, rather than using transfers (which are far less reliable).

So I think we should hold off on merging for now

source/dbo/Stored Procedures/Power BI Reporting/uspGenerateRecordOutcome.sql
```diff
@@ -48,7 +48,15 @@ BEGIN TRY
 			JOIN [$(NTBS)].ReferenceData.TbService tbs ON tbs.Code = te.TbServiceCode
 		WHERE TreatmentEventType = 'TransferOut';
 
-	WITH NotifyingServiceAndCode AS 
+	WITH NotifyingServiceAndCodeFromInitialEvent AS
+	(SELECT DISTINCT NotificationId,
```
JoshCadney, line 52: I think the `DISTINCT` here adds a false sense of security - if the notification did somehow end up with more than one treatment start/diagnosis made event then it's also possible that those events would have different TB services. I think it's better not to have the `DISTINCT` call and maybe comment saying each notification will only have one such event.

JoshCadney (changes requested):
Looks good, only a very minor improvement suggested

PR #141, opened 2022-01-07
NTBS-2973: Generate correct outcome for post mortem cases

LeoCie:
Generate correct outcomes for post mortem cases according to new rules.


JoshCadney (approved):
Looks good :-)

PR #142, opened 2022-01-07
NTBS-2987 Add SQL CMD variable for number of reporting years

JoshCadney:
Allow us to set the number of reporting years of data in the publish profile, so that we can change it on a per-environment basis


aptemus (approved):
Looks good

LeoCie (approved):
Looks good, does the local publish profile need to be updated too now?

aptemus:
> Looks good, does the local publish profile need to be updated too now?

Yes I think so - but that doesn't live in github does it?

LeoCie:
> > Looks good, does the local publish profile need to be updated too now?
> 
> Yes I think so - but that doesn't live in github does it?

I think the template does, it's called DEV-reporting.publish.xml. Isn't a huge issue anyway!

PR #143, opened 2022-01-13
NTBS-1526: initial view for service directory report

aptemus:
The service directory report will display information about regions and their staff + TB services and their staff, including contact details.  New view to expose the right data to Power BI.
It doesn't include the national team because their individual contact details have never been included in the service directory. I have put in a clause to exclude Softwire staff from being displayed because @JoshCadney has a London regional role which is really useful for testing but we don't want him to appear in the service directory.


LeoCie (approved):
Looks good! I'm guessing you'll split the view up into regions/services in power BI based on what is populated in this view as we have multiple rows per user?

JoshCadney (approved):
Looks good :-)

PR #145, opened 2022-01-14
NTBS-2997 provide descriptive periodic outcomes for reporting

aptemus:
The existing periodic outcomes in the reporting service are good for analysis but don't provide the same level of detail you can get from the notification overview.  To support a new report ('open cases') the users want to see 'Not evaluated - Still on treatment' for the 12 month outcome where applicable.

I've implemented this across the board, rather than just for the not evaluated outcome, in the expectation of further feature requests in the future.  I haven't implemented this for ETS records as it's not worth it at this stage. There are 38 open cases which aren't in NTBS, many of which will be migrated shortly.


JoshCadney (approved):
Looks good :-)

PR #146, opened 2022-01-18
NTBS-2981

aptemus:
Added 'notifying region code' so that we can ensure regional users can also access cases no longer in their region  but notified there.

I did it as a separate update because the calculation of the TB service code used to look up the region code is quite complicated.


LeoCie (approved):
Looks good!

PR #147, opened 2022-01-21
NTBS-2997 - add descriptive outcome for post-mortem cases

aptemus:
I realised when testing NTBS-2973 that the correct outcome for post-mortem cases is being set after `[uspGenerateReusableOutcomePeriodic]` is called for period 1, meaning the new 'descriptive' outcome fields are not set to the right value for post-mortem cases.

I added the logic used to return both the short and long outcome descriptions into` [uspUpdateOutcomesForPostMortemCases]`. It reproduces some of the logic in `ufnGetPeriodicOutcome` but I think it's the best way to handle it, given we make a special case for post-mortem case outcomes anyway.


JoshCadney (approved):
Looks good :-)

PR #148, opened 2022-01-24
NTBS-3005 add script to generate InitialSputumSmearResult and Initial…

jaice-dev:
Add script to fill InitialSputumSmearResult and InitialSputumPCRResult for Record_CaseData table.

Note: in all cases ‘sputum’ includes both sample types sputum - induced and sputum - spontaneous.

For each NTBS-sourced record in the report service, derive two new fields - Initial sputum smear result and Initial sputum PCR result.  The first will consider the test type ‘smear’ and the second the test type ‘PCR’ but otherwise the requirements are identical, as follows:

1. if there are no sputum results at all for the given test type, output ‘No result’
2. if there are only results of ‘Awaiting’, output ‘Awaiting’
3. if there are only results of ‘No result available', or only results of ‘No result available' and ‘Awaiting’, output ‘No result available’.
4. if there are results of ‘Positive’ and ‘Negative’ available, output the result on the earliest date on which a positive or negative was available
i) if both positive and negative are available on the earliest date, output ‘Positive’
ii) if only negative are available on the earliest date, output ‘Negative’.


source/ntbs-reporting.sqlproj
```diff
@@ -117,7 +117,6 @@
     <Folder Include="dbo\Views\Authorization" />
     <Folder Include="dbo\Views\Reference Data" />
     <Folder Include="dbo\Tables\Power BI Reporting" />
-    <Folder Include="dbo\Stored Procedures\Power BI Reporting" />
```
LeoCie, line 120: The changes to the project file don't look right to me, this shouldn't be removed

source/dbo/Tables/Power BI Reporting/Record_CaseData.sql
```diff
@@ -61,7 +61,9 @@
 	[DOTReceived] [varchar] (20) NULL,
 	[TestPerformed] [varchar] (10) NULL,
 	[SampleTaken] [varchar] (10) NULL,
-	[SputumResult] [varchar] (20) NULL,
+    [InitialSputumSmearResult] [varchar] (20) NULL,
+    [InitialSputumPCRResult] [varchar] (20) NULL,
+    [SputumResult] [varchar] (20) NULL,
```
LeoCie, line 66: Looks like the indentation is a little off with these new fields

source/dbo/Stored Procedures/Power BI Reporting/uspGenerateInitialSputumResults.sql
```diff
@@ -0,0 +1,79 @@
...
+
+	ResultRanking AS
+	(
+		SELECT 1 AS [Rank], 1 AS [SubRank], 'Positive' AS ResultName
+		UNION
+		SELECT 1 AS [Rank], 2 AS [SubRank], 'Negative' AS ResultName
+		UNION
+		SELECT 2 AS [Rank], NULL AS [SubRank], 'NoResultAvailable' AS ResultName
+		UNION
+		SELECT 3 AS [Rank], NULL AS [SubRank], 'Awaiting' AS ResultName
```
LeoCie, line 33: Very minor but seems weird to have some of the column names in brackets and some not

source/dbo/Stored Procedures/Power BI Reporting/uspGenerateInitialSputumResults.sql
```diff
@@ -0,0 +1,79 @@
...
+	),
+
+	NtbsInitialSputumSmearResults AS 
+	(
+		SELECT DISTINCT rr.NotificationId, rr.SourceSystem, FIRST_VALUE(mtr.[Result]) OVER (PARTITION BY rr.NotificationId ORDER BY rra.[Rank], mtr.TestDate, rra.[SubRank]) AS InitialSputumSmearResult
```
LeoCie, line 38: I would split each column into a new line for readability here and in the common table expression for PCR below

source/dbo/Stored Procedures/Power BI Reporting/uspGenerateInitialSputumResults.sql
```diff
@@ -0,0 +1,79 @@
...
+		FROM [test-ntbs].[dbo].[ManualTestResult] mtr
```
LeoCie, line 39: We shouldn't reference any specific databases in code, it would mean that a deployment to any environment that isn't our tests ones would fail. I think what you're looking for here is `[$(NTBS)].[dbo].[ManualTestResult]`.
`$(NTBS)` is a variable that is set by our deployment scripts to be the correct NTBS database we are deploying to. If this is confusing then let me know and we'll have a call to go through it

source/dbo/Stored Procedures/Power BI Reporting/uspGenerateReportingCaseData.sql
```diff
@@ -9,6 +9,7 @@ BEGIN TRY
 	--finally do stuff that makes sense just to do for all records, like the last recorded treatment outcome 
 
 	EXEC [dbo].[uspGenerateSputumResult]
+    EXEC [dbo].[uspGenerateInitialSputumResults]
```
LeoCie, line 12: Indentation different to the rest of the file here

source/dbo/Stored Procedures/Power BI Reporting/uspGenerateInitialSputumResults.sql
```diff
@@ -0,0 +1,79 @@
...
+		if (r only contains negative results):
+			'Negative'
+		else:
+			'Positive'
+	Else if (any results are 'No result available'):
+		'No Result available'
+	Else if (any results 'Awaiting'):
+		'Awaiting'
+	Else (no results):
+		'No result'
```
LeoCie, line 16: I'm not a fan of having pseudocode in comments, as they are usually to explain the code in words if the code doesn't explain itself (so if someone didn't understand code, they would still have trouble here)

LeoCie (changes requested):
Looks good! Changes are mainly cosmetic, and making sure we call the correct database

source/ntbs-reporting.sqlproj
jaice-dev: My IDE seems to have done a few strange things - Should be fixed now

source/dbo/Stored Procedures/Power BI Reporting/uspGenerateInitialSputumResults.sql
jaice-dev: I haven't ran this locally as the reporting service isn't set up for my local instance - I ended up testing it against a temporary version of Record_CaseData I made on Test, hence why I didn't need the variable

source/dbo/Stored Procedures/Power BI Reporting/uspGenerateInitialSputumResults.sql
```diff
@@ -0,0 +1,66 @@
...
+	)
+
+    UPDATE cd
+    SET cd.InitialSputumSmearResult =
+            (CASE sr.InitialSputumSmearResult
+                WHEN 'Positive' THEN 'Positive'
+                WHEN 'Negative' THEN 'Negative'
+                WHEN 'NoResultAvailable' THEN 'No result available'
+                WHEN 'Awaiting' THEN 'Awaiting'
+                ELSE 'No result'
```
LeoCie, line 52: Might be worth adding these to ResultRanking as display names, then you can select the display name instead of the result in the common table expressions and this turns into `CASE WHEN result <> NULL THEN result ELSE 'No result' END`

LeoCie (approved):
Looks good!

PR #150, opened 2022-01-26
NTBS-3016: Add post deploy script to add environment to database table

LeoCie:
No need for another variable, just putting the name of the NTBS database into a table so I can get it with Power BI and then map the environment -> notification URL

### DIFF
--- a/source/Publish Profiles/Pre-production/azure-load-test-reporting.publish.xml
+++ b/source/Publish Profiles/Pre-production/azure-load-test-reporting.publish.xml
@@ -102,5 +102,8 @@
     <SqlCmdVariable Include="ReportServer">
       <Value>ReportServer</Value>
     </SqlCmdVariable>
+    <SqlCmdVariable Include="NTBS_URL">
+      <Value>https://ntbs-test.e32846b1ddf0432eb63f.northeurope.aksapp.io/</Value>
+    </SqlCmdVariable>
   </ItemGroup>
 </Project>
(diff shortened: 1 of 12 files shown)


JoshCadney (commented):
This will work, but I'm not sure that it's better than adding an extra variable - it means adding logic to Power BI to map between database names and urls, which I think makes it more difficult to maintain (our philosophy has generally been: where there's a choice, put logic in the database, rather than Power BI).

What were your reasons for doing it this way?

aptemus (commented):
I agree with @JoshCadney - I can see why this is an attractive alternative to having to update all the publish profiles (even if you set a default value for a SQLCMD variable, you still have to put that value into each publish profile, it's v annoying).  I feel like it would be harder for another team to maintain, though.

source/Publish Profiles/Pre-production/azure-load-test-reporting.publish.xml
```diff
@@ -102,5 +102,8 @@
     <SqlCmdVariable Include="ReportServer">
       <Value>ReportServer</Value>
     </SqlCmdVariable>
+    <SqlCmdVariable Include="NTBS_URL">
+      <Value>https://ntbs-test.e32846b1ddf0432eb63f.northeurope.aksapp.io/</Value>
```
JoshCadney, line 106: This ought to be pointing at load test (not that it really matters, since load test doesn't have a reporting environment)

JoshCadney (approved):
Looks good :-)

PR #155, opened 2022-02-09
NTBS-3028 add notification to missing item report if referral date empty only for records after 2022

jordan-carlton:
To prevent every notification appearing on the missing items report, it has been decided that notifications should only show as missing 'referral received' if the notification is from 2022 or later.


JoshCadney (approved):
Looks good to me :-)

PR #157, opened 2022-02-10
NTBS-3042 add application usage view

jordan-carlton:
Add application usage view for internal dashboard report

### DIFF
--- a/source/dbo/Views/Power BI Reporting/vwApplicationUsage.sql
+++ b/source/dbo/Views/Power BI Reporting/vwApplicationUsage.sql
@@ -1,0 +1,22 @@
+﻿CREATE VIEW [dbo].[vwApplicationUsage]
+	AS
+    
+	WITH
+	Usage AS 
+	(
+        SELECT c.ISOYearWeek AS YearWeek, SUM(Q1.NumberOfRows) AS NumberOfAudits
+        FROM [Calendar] c
+            LEFT OUTER JOIN 
+                (SELECT CONVERT(DATE, [AuditDateTime]) AS AuditDate, COUNT(iD) AS NumberOfRows
+                FROM [$(NTBS_AUDIT)].[dbo].[AuditLogs]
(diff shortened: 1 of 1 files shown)


source/dbo/Views/Power BI Reporting/vwApplicationUsage.sql
```diff
@@ -0,0 +1,22 @@
+﻿CREATE VIEW [dbo].[vwApplicationUsage]
+	AS
+
+	WITH
+	Usage AS 
+	(
+		SELECT c.ISOYearWeek AS YearWeek, SUM(Q1.NumberOfRows) AS NumberOfAudits
+		FROM [Calendar] c
+			LEFT OUTER JOIN 
+			(SELECT CONVERT(DATE, [AuditDateTime]) AS AuditDate, COUNT(iD) AS NumberOfRows
```
LeoCie, line 10: `iD` here should be `ID`

source/dbo/Views/Power BI Reporting/vwApplicationUsage.sql
```diff
@@ -0,0 +1,22 @@
...
+			FROM [$(NTBS_AUDIT)].[dbo].[AuditLogs]
+			WHERE AuditUser != 'SYSTEM'
+			GROUP BY CONVERT(DATE, [AuditDateTime])) as q1 on q1.AuditDate = c.DateValue
+			WHERE c.DateValue between '2021-07-10' and GETUTCDATE()
```
LeoCie, line 14: There's a bit of a mix of case here. I know SQL doesn't mind about case but we should try to be consistent. I usually will use uppercase for keywords like you have with `GROUP BY` and such. So `as`, `on` and `and` should all be uppercased.

LeoCie (changes requested):
Just formatting changes, should always try to be consistent!

LeoCie (approved):
Looks good to me!

PR #159, opened 2022-02-28
NTBS-3071: Make test result summaries consistent

LeoCie:
Made the ranking a lookup table, as it is used in a few different places. 
Made the logic for all the different test result summaries consistent (with how we do it for the `Initial...` fields)

### DIFF
--- a/source/dbo/Stored Procedures/Power BI Reporting/uspGenerateNtbsCaseRecord.sql
+++ b/source/dbo/Stored Procedures/Power BI Reporting/uspGenerateNtbsCaseRecord.sql
@@ -10,17 +10,6 @@ BEGIN TRY
 		WHERE rr.SourceSystem = 'NTBS'
 		GROUP BY rr.NotificationId;
 
-	SELECT DISTINCT rr.NotificationId, ManualTestTypeId,
-		FIRST_VALUE(Result) OVER (PARTITION BY rr.NotificationId, mtr.ManualTestTypeId ORDER BY
-			CASE WHEN Result = 'Positive' THEN 1
-				WHEN Result = 'ConsistentWithTbOther' THEN 2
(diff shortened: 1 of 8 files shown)


source/dbo/Stored Procedures/Power BI Reporting/uspGenerateNtbsCaseRecord.sql
```diff
@@ -411,7 +379,6 @@ BEGIN TRY
 	WHERE rr.SourceSystem = 'NTBS'
 
 	DROP TABLE #TempDiseaseSites
-	DROP TABLE #TempManualTestResult
 	DROP TABLE #TempSocialContextVenues
 
 	--'Sample taken' should be set if there are any manually-entered test results which are NOT of type chest x-ray
```
LeoCie, line 384: I've spotted this comment too, should we be also removing Chest CT results when looking for SampleTaken?

JoshCadney: I assume that we should, but I'm not certain. @adilmirzaphe What do you think?

JoshCadney (approved):
Looks good :-)

source/dbo/Stored Procedures/Power BI Reporting/uspGenerateNtbsCaseRecord.sql
adilmirzaphe: @JoshCadney  vaguely recall this being discussed with Esther. Might be worth clarifying in catch up meeting tomorrow.

LeoCie: I'm going to put it in this ticket in JIRA, as it's a small change and still makes sense in this ticket.

PR #160, opened 2022-03-02
NTBS-3069 add world region to record_CaseData

jordan-carlton:
Add "world region" field to record_CaseData table for both NTBS and ETS records


JoshCadney (approved):
Looks good :-)

PR #163, opened 2022-03-14
NTBS-2943 Tidy up reporting code

jordan-carlton:
Tidy up reporting code:

There were 4 folders called 'Power BI Reporting' in the Functions, Stored Procedures, Tables and Views folders.

I have left the Functions folder as is because some of these are not being used (I'm reticent to delete them just in case they're being used in Power BI somewhere), and others are being used in multiple views/tables making them hard to categorise.

I have left a Power BI Reporting folder in the Views table as the 5 views inside didn't fit neatly into any other category. Do let me know if you have ideas here.


JoshCadney (approved):
Looks good :-)

PR #167, opened 2022-03-17
NTBS-3052

jordan-carlton:
Add unmatched specimens and unmatched specimens linked notifications tables


source/dbo/Stored Procedures/Unmatched Specimens/uspGenerateUnmatchedSpecimensLinkedNotifications.sql
```diff
@@ -0,0 +1,66 @@
...
+	SELECT us.ReferenceLaboratoryNumber, nsm.NotificationID, 'RejectPotentialMatch' AS EventType INTO #PreviousPossibleMatch
+	FROM UnmatchedSpecimens us
+	JOIN [$(NTBS_Specimen_Matching)].dbo.NotificationSpecimenMatch nsm ON nsm.ReferenceLaboratoryNumber = us.ReferenceLaboratoryNumber
+	WHERE nsm.MatchType = 'Rejected-Possible'
+
+	SELECT mr.clientsourceID AS ReferenceLaboratoryNumber, mr.rclientsourceID AS NotificationID, 'PoorQualityMatch' AS EventType INTO #PoorQualityMatch
+	FROM UnmatchedSpecimens us
+	JOIN [$(NTBS_Specimen_Matching)].dbo.MatchResults mr ON mr.clientsourceID = us.ReferenceLaboratoryNumber
+	--Exclude any matches that have previously been rejected (otherwise every rejection has an accompanying match)
+	LEFT JOIN #PreviousMatch pm on pm.NotificationID = mr.rclientsourceID AND pm.ReferenceLaboratoryNumber = mr.clientsourceID
```
jordan-carlton, line 24: This isn't strictly necessary, but it prevents any rejected match from showing as both a rejection and a poor quality match in the report

source/dbo/Stored Procedures/Unmatched Specimens/uspGenerateUnmatchedSpecimensLinkedNotifications.sql
```diff
@@ -0,0 +1,66 @@
...
+	LEFT JOIN #PreviousPossibleMatch ppm on ppm.NotificationID = mr.rclientsourceID AND ppm.ReferenceLaboratoryNumber = mr.clientsourceID
+	WHERE pm.EventType IS NULL AND ppm.EventType IS NULL
+	AND mr.[weight] < 34.50
```
jordan-carlton, line 27: I think the logic around poor quality matches is more complicated than this (there are multiple confirmed matches with weights in the 20's) but I think this will do

source/dbo/Stored Procedures/Unmatched Specimens/uspGenerateRejectSpecimenUsers.sql
```diff
@@ -0,0 +1,30 @@
+﻿CREATE PROCEDURE [dbo].[uspGenerateRejectSpecimenUsers]
+AS
+BEGIN TRY
+
+	TRUNCATE TABLE [dbo].[RejectSpecimenUsers]
```
JoshCadney, line 5: I think `RejectedSpecimen` would be a better name for this table

source/dbo/Stored Procedures/Unmatched Specimens/uspGenerateUnmatchedSpecimens.sql
```diff
@@ -0,0 +1,84 @@
...
+	INSERT INTO #MatchTypeRanking VALUES (2, 'Rejected')
+
+	SELECT ranked.ReferenceLaboratoryNumber INTO #AllMatchesRejected FROM (
+		SELECT DISTINCT nsm.ReferenceLaboratoryNumber, FIRST_VALUE(mtr.ResultName) OVER (PARTITION BY nsm.ReferenceLaboratoryNumber ORDER BY mtr.[rank] ASC) AS [Result]
+			FROM [$(NTBS_Specimen_Matching)].dbo.NotificationSpecimenMatch nsm
+			INNER JOIN [$(NTBS_Specimen_Matching)].dbo.LabSpecimen ls ON ls.ReferenceLaboratoryNumber = nsm.ReferenceLaboratoryNumber
+			INNER JOIN vwNotificationYear ny ON ny.NotificationYear = YEAR(ls.SpecimenDate)
+			INNER JOIN #MatchTypeRanking mtr ON mtr.ResultName = nsm.MatchType
+			) ranked
+		WHERE ranked.Result = 'Rejected' OR ranked.Result = 'Rejected-Possible'
```
JoshCadney, line 28: I think you could use `FIRST_VALUE(mtr.[rank])` in the select and then `WHERE ranked.Rank = 2` here to simplify it a little. 

I think at the moment you could get duplicates here, if there was both a "Rejected" and a "Rejected-Possible" specimen (though I don't know if that's actually possible). But the change suggested above would remove those duplicates if they can occur.

source/dbo/Stored Procedures/Unmatched Specimens/uspGenerateUnmatchedSpecimens.sql
```diff
@@ -0,0 +1,84 @@
...
+			,ls.[EarliestRecordDate]
+			,COALESCE(phec2.[Code], phec1.[Code]) AS RegionCode
+			,COALESCE(phec2.[Name], phec1.[Name]) AS Region
+		FROM #UnmatchedSpecimens us
+		JOIN [$(NTBS_Specimen_Matching)].dbo.LabSpecimen ls ON ls.ReferenceLaboratoryNumber = us.ReferenceLaboratoryNumber
+		LEFT JOIN [$(NTBS_Specimen_Matching)].dbo.LabHospitalRegionMapping mapping ON mapping.LaboratoryName = ls.LaboratoryName
+		LEFT JOIN [$(NTBS)].ReferenceData.PHEC phec1 ON phec1.Code = mapping.RegionCode
+		LEFT JOIN [$(NTBS)].ReferenceData.Hospital h ON h.HospitalId = mapping.HospitalId
+		LEFT JOIN [$(NTBS)].ReferenceData.TbService tbs ON tbs.Code = h.TBServiceCode
+		LEFT JOIN [$(NTBS)].ReferenceData.PHEC phec2 ON phec2.Code = tbs.PHECCode
```
JoshCadney, line 71: Why do we need `phec2`? i.e. surely if the Hospital appears in the mapping table then so does the PHEC?

source/dbo/Stored Procedures/Unmatched Specimens/uspGenerateUnmatchedSpecimens.sql
```diff
@@ -0,0 +1,84 @@
...
+		LEFT JOIN [$(NTBS)].ReferenceData.Hospital h ON h.HospitalId = mapping.HospitalId
+		LEFT JOIN [$(NTBS)].ReferenceData.TbService tbs ON tbs.Code = h.TBServiceCode
+		LEFT JOIN [$(NTBS)].ReferenceData.PHEC phec2 ON phec2.Code = tbs.PHECCode 
+
+	EXEC dbo.uspGenerateUnmatchedSpecimensLinkedNotifications
+
+	DROP TABLE #NoMatchesFound
+	DROP TABLE #MatchTypeRanking
+	DROP TABLE #AllMatchesRejected
+	DROP TABLE #UnmatchedSpecimens
```
JoshCadney, line 78: I presume we're not using these temporary tables in the stored proc that we're executing above? If so, then we should move the `DROP TABLE` statements above.

source/dbo/Stored Procedures/Unmatched Specimens/uspGenerateUnmatchedSpecimensLinkedNotifications.sql
```diff
@@ -0,0 +1,66 @@
...
+
+	CREATE TABLE #AllMatches (ReferenceLaboratoryNumber NVARCHAR(50), NotificationID int, EventType NVARCHAR(50))
+	INSERT INTO #AllMatches SELECT * FROM #PreviousMatch
+	INSERT INTO #AllMatches SELECT * FROM #PreviousPossibleMatch
+	INSERT INTO #AllMatches SELECT * FROM #PoorQualityMatch
```
JoshCadney, line 32: Given that these aren't matches, this seems like a fairly confusing name. What about `AllLinks`?

source/dbo/Stored Procedures/Unmatched Specimens/uspGenerateUnmatchedSpecimensLinkedNotifications.sql
```diff
@@ -0,0 +1,66 @@
...
+
+	INSERT INTO [dbo].[UnmatchedSpecimensLinkedNotifications]
+	SELECT am.ReferenceLaboratoryNumber
+		,CASE
+			WHEN am.EventType = 'RejectPotentialMatch' THEN 'Possible match rejected by ' + u.UserDisplayName + ' on ' + FORMAT(u.RejectionTime, 'dd MMM yyyy')
+			WHEN am.EventType = 'Unmatch' THEN 'Confirmed match rejected by ' + u.UserDisplayName + ' on ' + FORMAT(u.RejectionTime, 'dd MMM yyyy')
```
JoshCadney, line 38: It would be better to allow Power BI to format these dates so that they are consistent with the other reporting dates.

How difficult would it be to move this logic into Power BI? (If it's going to be hard, then I'm happy for this to be a "nice to have" that we don't have time for)

source/dbo/Stored Procedures/Unmatched Specimens/uspGenerateUnmatchedSpecimensLinkedNotifications.sql
```diff
@@ -0,0 +1,66 @@
...
+			WHEN am.EventType = 'PoorQualityMatch' THEN 'Poor quality potential match'
+		END AS NotificationLinkReason
+		,am.NotificationID
+		,n.NotificationStatus
+		,p.NhsNumber
+		,p.Dob AS BirthDate
+		,UPPER(p.FamilyName) + ', ' + p.GivenName AS [Name]
```
JoshCadney, line 45: Is this logic that is usually done at the database level?

source/dbo/Tables/Report Unmatched Specimens/UnmatchedSpecimens.sql
```diff
@@ -0,0 +1,26 @@
+﻿CREATE TABLE [dbo].[UnmatchedSpecimens]
+(
+	[ReferenceLaboratoryNumber] NVARCHAR(50) NOT NULL PRIMARY KEY,
+	[SpecimenDate] DATETIME,
+	[SampleType] NVARCHAR(255),
+	[LaboratoryName] NVARCHAR(65),
+	[ReferenceLaboratory] NVARCHAR(65),
+    [Species] VARCHAR(30),
```
JoshCadney, line 8: It looks like something weird may be happening with the whitespace in this file - a mix of tabs and spaces perhaps?

JoshCadney (changes requested):
Overall looks really good :-)

source/dbo/Stored Procedures/Unmatched Specimens/uspGenerateUnmatchedSpecimensLinkedNotifications.sql
jordan-carlton: It's the value of 'AutoMatch' from the MatchingConfiguration table in the SpecimenMatching db

source/dbo/Stored Procedures/Unmatched Specimens/uspGenerateUnmatchedSpecimensLinkedNotifications.sql
jordan-carlton: I think it would be fairly difficult to do as is - you'd have to split the string, reformat it and then concatenate them back together. It would probably be easier to create a new RejectionDate field for the table and do the concatenation that way in powerBI, (It will also simplify the logic around selecting which rows to apply the concatenation to ie rows where RejectionDate is not null).

I think this probably is worth doing!

source/dbo/Stored Procedures/Unmatched Specimens/uspGenerateUnmatchedSpecimensLinkedNotifications.sql
jordan-carlton: I'll move this into powerBI

source/dbo/Stored Procedures/Unmatched Specimens/uspGenerateUnmatchedSpecimensLinkedNotifications.sql
JoshCadney: In that case we should fetch the value directly from the table - we shouldn't be expecting someone who changes the configured value to have to remember to change this too!

source/dbo/Stored Procedures/Unmatched Specimens/uspGenerateUnmatchedSpecimensLinkedNotifications.sql
```diff
@@ -23,42 +23,45 @@ BEGIN TRY
 	--Exclude any matches that have previously been rejected (otherwise every rejection has an accompanying match)
 	LEFT JOIN #PreviousMatch pm on pm.NotificationID = mr.rclientsourceID AND pm.ReferenceLaboratoryNumber = mr.clientsourceID
 	LEFT JOIN #PreviousPossibleMatch ppm on ppm.NotificationID = mr.rclientsourceID AND ppm.ReferenceLaboratoryNumber = mr.clientsourceID
+	JOIN [$(NTBS_Specimen_Matching)].dbo.MatchingConfiguration mc ON mr.[weight] < mc.AutoMatch
```
JoshCadney, line 26: Doing this as a `JOIN` is semantically strange. I think this code will be easier to follow if you select the value into a variable, and then use a `WHERE` clause as you were doing before.

source/dbo/Stored Procedures/Unmatched Specimens/uspGenerateUnmatchedSpecimensLinkedNotifications.sql
```diff
@@ -23,42 +23,45 @@ BEGIN TRY
...
 	INSERT INTO [dbo].[UnmatchedSpecimensLinkedNotifications]
-	SELECT am.ReferenceLaboratoryNumber
+	SELECT al.ReferenceLaboratoryNumber
 		,CASE
-			WHEN am.EventType = 'RejectPotentialMatch' THEN 'Possible match rejected by ' + u.UserDisplayName + ' on ' + FORMAT(u.RejectionTime, 'dd MMM yyyy')
-			WHEN am.EventType = 'Unmatch' THEN 'Confirmed match rejected by ' + u.UserDisplayName + ' on ' + FORMAT(u.RejectionTime, 'dd MMM yyyy')
-			WHEN am.EventType = 'PoorQualityMatch' THEN 'Poor quality potential match'
+			WHEN al.EventType = 'RejectPotentialMatch' THEN 'Possible match rejected by ' + rs.UserDisplayName + ' on '
+			WHEN al.EventType = 'Unmatch' THEN 'Confirmed match rejected by ' + rs.UserDisplayName + ' on '
+			WHEN al.EventType = 'PoorQualityMatch' THEN 'Poor quality potential match'
```
JoshCadney, line 40: I think it would be better to keep the type and the date in the table, and then do all of the formatting in Power BI, rather than half here and half there.

JoshCadney (changes requested):

JoshCadney (approved):
Looks good :-)

PR #170, opened 2022-03-23
NTBS-3052 Unmatched Specimens Report

jordan-carlton:
Remove primary key from Linked Notifications table following QA feedback

### DIFF
--- a/source/dbo/Tables/Report Unmatched Specimens/UnmatchedSpecimensLinkedNotifications.sql
+++ b/source/dbo/Tables/Report Unmatched Specimens/UnmatchedSpecimensLinkedNotifications.sql
@@ -1,6 +1,6 @@
 ﻿CREATE TABLE [dbo].[UnmatchedSpecimensLinkedNotifications]
 (
-	[ReferenceLaboratoryNumber] NVARCHAR(50) NOT NULL PRIMARY KEY,
+	[ReferenceLaboratoryNumber] NVARCHAR(50) NOT NULL,
 	[NotificationLinkReason] NVARCHAR(255),
 	[RejectionDate] DATETIME2(7),
 	[User] NVARCHAR(256),


source/dbo/Tables/Report Unmatched Specimens/UnmatchedSpecimensLinkedNotifications.sql
```diff
@@ -1,17 +1,16 @@
...
-	[NotificationStatus] NVARCHAR(30),
-	[NhsNumber] NVARCHAR(10),
-	[BirthDate] DATETIME2(7),
-	[FamilyName] NVARCHAR(35),
-	[GivenName] NVARCHAR(35),
-	[Sex] NVARCHAR (200),
-	[Address] NVARCHAR (150),
-	[Postcode] NVARCHAR(10),
-	[RegionCode] NVARCHAR(50)
+	[Denotified] NVARCHAR(30),
```
LeoCie, line 8: Is there supposed to be a change to the proc that adds to this table as part of this commit? I don't think denotified is a good name if it's still going to contain the status.

source/dbo/Tables/Report Unmatched Specimens/UnmatchedSpecimensLinkedNotifications.sql
```diff
@@ -1,17 +1,16 @@
 ﻿CREATE TABLE [dbo].[UnmatchedSpecimensLinkedNotifications]
 (
-	[ReferenceLaboratoryNumber] NVARCHAR(50) NOT NULL PRIMARY KEY,
+	[ReferenceLaboratoryNumber] NVARCHAR(50) NOT NULL,
```
JoshCadney, line 3: Should we be replacing this with a primary key on the `ReferenceLaboratoryNumber` and the `NotificationID`?

Perhaps we shouldn't, as it's at least conceptually possible that the same notification could be related to a lab result in two different ways

JoshCadney (changes requested):
No mark-ups from me, but it looks like Leon's point needs addressing

source/dbo/Tables/Report Unmatched Specimens/UnmatchedSpecimensLinkedNotifications.sql
jordan-carlton: I didn't mean to include these changes (they were from a change I made to not join to NTBS directly, but then discarded)

LeoCie (approved):
Looks good now!

PR #226, opened 2026-06-17
Update reporting  to incoporate asylum seeker ranges.

seanVincent9:
Tables - Record case data and create forest extract table Stored proceudres
Populate forest
NTBS case record.


adilmirzaphe (approved):
Approved now as will be included for next release

